PR #2, opened 2016-08-22
Edu benefits json schema

### LINKED ISSUE #1: add edu benefits json schema

lihanli:
closes #1


lihanli:
@jbalboni ok i made some changes, let me know if this looks ok

src/common/options-for-select.js
```diff
@@ -0,0 +1,1594 @@
...
+  6: thirtyDays,
+  7: thirtyOneDays,
+  8: thirtyOneDays,
+  9: thirtyDays,
+  10: thirtyOneDays,
+  11: thirtyDays,
+  12: thirtyOneDays
+};
+
+const vaMedicalFacilities = {
```
ayaleloehr: No action here, but just wanted to mention how awesome would it be to have this list a single time across all of our applications instead of individually for every product!

src/edu-benefits/schema.js
```diff
@@ -0,0 +1,436 @@
+// Relocate options-for-select.
+import options from '../common/options-for-select';
```
jbalboni: Now that we're in a separate repo, should we change the name of options-for-select? It doesn't make much sense in here. That file has static data that could be used in a number of contexts. We could probably change it in vets-website, too.

src/edu-benefits/schema.js
```diff
@@ -0,0 +1,436 @@
...
+        properties: {
+          name: {
+            type: 'string',
+            'enum': [
+              'comensation',
+              'pension',
+              'vocationalRehab',
+              'homeLoan',
+              'insurance',
+              'educationBenefits.chapter30',
```
jbalboni: What's the namespacing in these fields for?

.nvmrc
```diff
@@ -0,0 +1 @@
+6.4.0
```
jbalboni: The other repos I've seen here seem to be on Node 4. I don't know if it's important to stay on that version or if we can go to 6.

src/edu-benefits/schema.js
lihanli: they're all education benefits forms

src/edu-benefits/schema.js
lihanli: @jbalboni i renamed it to constants, any other ideas or is that good

jbalboni:
LGTM

PR #7, opened 2016-08-26
Set additionalProperties: false in edu-benefits

markolson:
`additionalProperties: false` is a middle ground between having to explicitly set everything as `optional` or `required`, letting us verify we don't have any unaccounted-for data being submitted.


markolson:
Added this because, as I work on validating my sample data in vets-api, `:strict` was requiring that every field be present which is ... not correct.

PR #10, opened 2016-08-30
Update to match proposed question flow

lihanli:


src/edu-benefits/schema.js
```diff
@@ -341,89 +327,106 @@ module.exports = {
...
+        type: 'object',
+        properties: {
+          name: {
+            type: 'string'
+          },
+          city: {
+            type: 'string'
+          },
+          state: {
+            type: 'string'
```
jbalboni: Should we use an enum here so we only allow valid states?

markolson:
For reference, all my comments will be based off of the contents of the `edu-benefits-team/Discovery/Spool file sample2.docx` examples.

src/edu-benefits/schema.js
```diff
@@ -286,48 +262,58 @@ module.exports = {
       }
     },
     currentlyActiveDuty: {
-      type: 'boolean'
-    },
-    terminalLeaveBeforeDischarge: {
-      type: 'boolean'
+      type: 'object',
```
jbalboni: If this grouping makes things easier on the back end I'm fine with it, but having the fields as individual boolean fields also works for the front end.

dist/edu-benefits-schema.json
```diff
@@ -365,62 +381,25 @@
         }
       }
     },
-    "previouslyFiledClaimWithVa": {
```
markolson: These "previous claim" fields exist in the Spool file - we should follow up and see if they need to be collected.

dist/edu-benefits-schema.json
```diff
@@ -320,19 +324,31 @@
...
+      ]
+    },
+    "secondaryContact": {
       "type": "object",
       "properties": {
         "fullName": {
           "$ref": "#/definitions/fullName"
         },
-        "sameAddressAndPhone": {
+        "sameAddress": {
```
markolson: This is output as "Same address and phone as claimant" in the spool file, and may be an important distinction. Maybe not. So another thing to follow up on.

src/edu-benefits/schema.js
lihanli: i decided to group them together because on the question flow they are dependent questions to someone answering yes to currently active duty

src/edu-benefits/schema.js
lihanli: ok i added it

dist/edu-benefits-schema.json
```diff
@@ -504,25 +499,56 @@
...
+              "enum": [
+                "chapter33",
+                "chapter30",
+                "chapter1606",
+                "chapter32"
+              ]
+            }
+          },
+          "involuntarilyCalledToDuty": {
+            "type": "boolean"
```
jbalboni: On the front end, this has three options: Yes, No, and Not Applicable. Should see if we need that third option, probably.

lihanli: i checked the form and saw n/a so i added it

jbalboni:
It looks like HCA was using 'Y' and 'N' as the values for Yes/No radio buttons. In some places here we're using booleans for those questions and for involuntarilyCalledToDuty we have 'yes', 'no', and 'n/a'.

I think 'Y' and 'N' work better for the data structure of the form on the front end, but I don't think I would want to do that in the json schema. @markolson @lihanli, thoughts?

jbalboni:
Also, I'm fine with merging this at any time. I'm sure I'll have other questions as we're building the form, but I can open issues for them.

markolson:
Agreed. I'd like to have the frontend starting to use this in their tests, so let's basically support whatever they're generating for now and move to get that unified with VONAPP/Spoolfile as needed.

markolson:
As for Yes/No/Not Applicable, I have no idea! The sample files we got have N/A in a few places, and in others have the fields left blank, so it's a bit of a mix. I'm working on a email to ask lots of little clarification questions and will include this as well.

lihanli:
i think i would want to keep yes/no as boolean in the json schema unless we need to support n/a

PR #12, opened 2016-09-02
Rename keys to hca ones

### LINKED ISSUE #11: Consistent naming
@markolson @lihanli I noticed that some of the names of the fields are not consistent with the names HCA uses for the same fields (e.g., for the veteran's name, HCA uses `veteranFullName` and the education schema just uses `fullName`). While the veteran objects are separate now, I think eventually we may move to using a single veteran object for all of vets.gov, so we might as well start making things consistent now. Thoughts?

lihanli:
closes #11

### DIFF
--- a/dist/edu-benefits-schema.json
+++ b/dist/edu-benefits-schema.json
@@ -321,7 +321,7 @@
     "veteranSocialSecurityNumber": {
       "$ref": "#/definitions/ssn"
     },
-    "address": {
+    "veteranAddress": {
       "$ref": "#/definitions/address"
     },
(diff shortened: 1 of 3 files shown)


saneshark:
Just going to throw this out there -- since we're going to be using JSON API and want the vets-api to be consistent, it's worth noting that HCA F/E will most likely need to be remapped to use the format that JSON API uses for JSON anyway. 

I'm only making this as a suggestion, but HCA is likely to need changes anyway.

akainic:
LGTM!

PR #20, opened 2016-09-15
Add back previous claims

jbalboni:
Previous claims is back in the wireframes, so I've pulled what was there earlier and tweaked it to fit the wireframed pages.

The list of claim types is from the fields spreadsheet.


jbalboni:
@lihanli Can you take a look?

PR #23, opened 2016-09-23
Change FAA field back to string

jbalboni:
The wireframe has changed back to having FAA certificates as a textarea, so this should be a string again.


jbalboni:
@lihanli Can you take a look?

PR #24, opened 2016-09-28
Address field updates

jbalboni:
We're simplifying the address field into the same set of fields regardless of country. 

Now we have:

street
street2 (not used on front end right now)
city
state
postalCode

I've also limited postal code to 10 characters for USA/CAN/MEX, since that's the max for any of those countries.


jbalboni:
@lihanli Can you take a look?

cc @markolson

PR #25, opened 2016-09-28
Add extra states for military bases

jbalboni:
I missed this in the last PR, we need to allow the military state codes for military base addresses.


jbalboni:
@lihanli Can you take a look?

PR #27, opened 2016-10-05
Remove some required fields

jbalboni:
There are some required fields in here that are no longer required on the front end.


jbalboni:
@lihanli Can you take a look?

PR #29, opened 2016-10-12
Update edu benefits apply to question

jbalboni:
One of the questions for the tours of duty is changing to be opt out with a textarea for the explanation.

It looks like we were missing the question before (because you had to make a choice if you checked it), so I've added it.

`applyPeriodToSelected` is going to be true by default, and the user can fill in benefitsToApplyTo if they uncheck it.


jbalboni:
@markolson @lihanli FYI

PR #31, opened 2016-10-17
allow phone numbers with dashes

lihanli:


dist/edu-benefits-schema.json
```diff
@@ -278,7 +278,7 @@
     },
     "phone": {
       "type": "string",
-      "pattern": "^[0-9]{10}$"
+      "minLength": 10
```
jbalboni: We don't want to restrict the characters at all?

lihanli:
i changed it to allow only digits and dashes

PR #33, opened 2016-10-21
Make recent form updates

jbalboni:
These are the recent updates for the form changes we've made

- Previous claims are removed
- Hours is changed to a number
- Involuntarily called question is removed

### DIFF
--- a/dist/edu-benefits-schema.json
+++ b/dist/edu-benefits-schema.json
@@ -382,52 +382,6 @@
         }
       }
     },
-    "previouslyFiledClaimWithVa": {
-      "type": "boolean"
-    },
-    "previousVaClaims": {
-      "type": "array",
-      "items": {
-        "type": "object",
-        "properties": {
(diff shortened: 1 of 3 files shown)


jbalboni:
@lihanli Can you take a look?

dist/edu-benefits-schema.json
```diff
@@ -662,7 +606,7 @@
             "$ref": "#/definitions/dateRange"
           },
           "hours": {
-            "type": "integer"
+            "type": "number"
```
akainic: @jbalboni I thought the outcome for this was to prevent floats on the front-end, not allow them in the schema?

jbalboni: Hmm, maybe I misunderstood. I thought the issue was that we wanted them to match and for the front end to not send NaN or null.

We can definitely change it, though do we want to be more restrictive? I think decimals could be valid for coursework hours.

PR #34, opened 2016-10-28
Remove commissionYear as required in edu-benefits

markolson:


jbalboni:
@markolson Need to run `npm run build`.

src/edu-benefits/schema.js
```diff
@@ -280,7 +280,7 @@ module.exports = {
           }
         }
       },
-      required: ['commissionYear', 'rotcScholarshipAmounts']
+      required: ['rotcScholarshipAmounts']
```
cweber: ROTC Scholarship shouldn't be required either. Am I reading that correctly?

Yes/no exposes it and indicates, but we (unintuitively [<-- not a word?]) don't want to use that to require any fields.

jbalboni: This won't force any required fields, it just means the scholarship list has to be there. By default, the front end will send back an empty list.

markolson: Yep. `required` at each object's level only means that they're required at that level, and `seniorRotc` isn't set as required at the root level, so it can be left off.

PR #35, opened 2016-10-31
change months to number instead of integer to allow for floats

akainic:
Fixes https://github.com/department-of-veterans-affairs/vets-api/issues/283

@jbalboni or @lihanli can you please review?

### DIFF
--- a/dist/edu-benefits-schema.json
+++ b/dist/edu-benefits-schema.json
@@ -633,7 +633,7 @@
             "type": "string"
           },
           "months": {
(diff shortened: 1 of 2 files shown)


jbalboni:
Weird, I thought we did this? 

@akainic You'll need to run `npm run build` to generate the schema, and then make sure the tests pass `npm run test`.

akainic:
@jbalboni guess it was never completed? and thanks for those additional instructions! Just did that, let me know if that looks okay now.

PR #38, opened 2016-11-14
Add hca

lihanli:
closes department-of-veterans-affairs/healthcare-application-team#503


jbalboni:
@lihanli Are there any differences between the output in the dist folder here and what's in healthcare-application now?

akainic:
I think there were still some outstanding questions on the json schema, so let's just hold for a moment on merging this until we can answer those questions if that's okay. Specifically, how we're handling the dependencies for financial disclosure.

lihanli:
the diff is these states have been added `"AA", "AE", "AP"`

PR #42, opened 2016-11-28
max 50 chars for insurancePolicyHolderName

lihanli:


akainic:
Looks great! Let's try to sync merging this with the front-end change (I will address this today) so as to avoid a situation where a veteran would pass our validations but fail on submit. I'll update this PR once the front-end change is ready.

PR #46, opened 2016-12-14
Add privacy policy field to schema for education and HCA

akainic:
@markolson this won't cause a submission to fail if this field doesn't exist, correct?


markolson:
No, this just _allows_ that field to exist, as we have some strict checking enabled.

akainic:
Great :)

PR #48, opened 2016-12-15
Require the privacy flag to be true

markolson:
This change will cause our controllers to submissions that have the `privacyAgreementAccepted=false` as invalidly formatted submissions. Tests will be written in each application to ensure that this behavior isn't changed in the future.


jcmeloni:
LGTM

PR #49, opened 2017-01-12
Form 1995

lihanli:
https://github.com/department-of-veterans-affairs/vets-website/issues/4551

### DIFF
--- a/src/change-of-program/schema.js
+++ b/src/change-of-program/schema.js
@@ -13,7 +13,9 @@ module.exports = {
     'ssn',
     'school',
     'bankAccount',
-    'serviceBefore1977'
+    'serviceBefore1977',
+    'date',
+    'dateRange'
   ]),
   properties: {
     veteranFullName: {
@@ -85,21 +87,22 @@ module.exports = {
       items: {
         type: 'object',
         properties: {
+          serviceBranch: {
+            type: 'string'
+          },
           dateRange: {
             $ref: '#/definitions/dateRange'
           },
-          serviceBranch: {
-            type: 'string'
+          involuntarilyCalledToDuty: {
+            type: 'boolean'
           },
           serviceStatus: {
             type: 'string'
           },
-          applyPeriodToSelected: {
-            type: 'boolean'
-          },
-          benefitsToApplyTo: {
-            type: 'string'
-          },
+          nationalGuardType: {
+            type: 'string',
+            enum: ['title10', 'title32']
+          }
         },
         required: ['dateRange', 'serviceBranch']
       }
--- a/src/change-of-program/schema.js
+++ b/src/change-of-program/schema.js
@@ -10,7 +10,8 @@ module.exports = {
     'fullName',
     'address',
     'phone',
-    'ssn'
+    'ssn',
+    'school'
   ]),
   properties: {
     veteranFullName: {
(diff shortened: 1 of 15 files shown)


src/change-of-program/schema.js
```diff
@@ -0,0 +1,120 @@
+import definitions from '../common/definitions';
+import _ from 'lodash';
+
+module.exports = {
```
jbalboni: If you're using ES6 imports here, could consider using ES6 export syntax, instead of commonjs.

src/change-of-program/schema.js
```diff
@@ -0,0 +1,120 @@
...
+    'dateRange'
+  ]),
+  properties: {
+    veteranFullName: {
+      $ref: '#/definitions/fullName'
+    },
+    veteranAddress: {
+      $ref: '#/definitions/address'
+    },
+    dayPhone: {
```
jbalboni: I think we had home and mobile phone before. Should we keep them the same?

src/change-of-program/schema.js
```diff
@@ -0,0 +1,120 @@
...
+      type: 'string'
+    },
+    email: {
+      type: 'string',
+      format: 'email'
+    },
+    veteranSocialSecurityNumber: {
+      $ref: '#/definitions/ssn'
+    },
+    benefit: {
```
jbalboni: I think you can pick more than one of these.

src/change-of-program/schema.js
```diff
@@ -0,0 +1,120 @@
...
+      type: 'string',
+      enum: ['chapter33', 'chapter30', 'chapter32', 'chapter1606', 'chapter1607', 'transferOfEntitlement']
+    },
+    educationType: {
```
jbalboni: I think on the form this will be a dropdown and only have one value.

src/change-of-program/schema.js
```diff
@@ -0,0 +1,120 @@
...
+    programName: {
+      type: 'string'
+    },
+    newSchool: {
+      $ref: '#/definitions/school'
+    },
+    oldSchool: {
+      $ref: '#/definitions/school'
+    },
+    reasonForChange: {
```
jbalboni: There's also a stop training date, I believe.

src/change-of-program/schema.js
```diff
@@ -0,0 +1,120 @@
...
+      items: {
+        type: 'object',
+        properties: {
+          serviceBranch: {
+            type: 'string'
+          },
+          dateRange: {
+            $ref: '#/definitions/dateRange'
+          },
+          involuntarilyCalledToDuty: {
```
jbalboni: I think service periods just has branch and date range.

src/change-of-program/schema.js
```diff
@@ -0,0 +1,120 @@
...
+    },
+    oldSchool: {
+      $ref: '#/definitions/school'
+    },
+    reasonForChange: {
+      type: 'string'
+    },
+    bankAccountChange: {
+      type: 'string',
+      enum: ['startOrChange', 'stop']
```
jbalboni: I read this as having three options

jbalboni:
@lihanli Can you bump the version in package.json? We need to do this each time we make a change, otherwise npm doesn't pull in new changes correctly in vets-website.

src/change-of-program/schema.js
lihanli: on this form updated dec 2016 it says select only one http://www.vba.va.gov/pubs/forms/vba-22-1995-are.pdf

markolson: @cweber @goldenmeanie can either of you weigh in on this?

cweber: This form is for modifying your existing benefit. Select one should be accurate.

PR #50, opened 2017-01-12
Adding 'Other' as 'stateOfBirth' option

knkski:
Need to add an 'Other' state option, but don't want it to conflict with other apps using the schema, so adding an additional field


src/healthcare-application/schema.js
```diff
@@ -207,6 +207,9 @@ module.exports = {
     stateOfBirth: {
       'enum': states
     },
+    placeOfBirthState: {
+      'enum': states + [{ label: 'Other', value: 'Other' }]
```
markolson: Ah, I see re: https://github.com/department-of-veterans-affairs/vets-api/pull/696/commits/d0553cc88087173d1c9fec5e8fbda000c80f827c#r95842564

What I'd meant in our chat was that `states + [{ label: 'Other', value: 'Other' }]` could be the enum for `stateOfBirth`. I just wanted to avoid redefining `states` itself.

jbalboni:
@knkski Need to bump the package.json version when we make a change, too.

PR #53, opened 2017-02-02
Tweaks for 1995 front end

jbalboni:
These are the schema changes I made for the 1995 front end. Some of them affect common components. I don't think they'll have a significant impact on the backend, but let me know if they do.

Changes:

- Removed `present` as option for end of date range. We don't ever use this.
- Removed `salutation` from name. Also not ever used.
- Added a pattern for file number, since we're validating this on the front end.
- Added `type` to enum fields, since this is required for the front end.
- Changed the order of some fields/enums because it affects the front end.
- Added preferredContactMethod field (same as 1990 form).

### DIFF
--- a/dist/change-of-program-schema.json
+++ b/dist/change-of-program-schema.json
@@ -7,9 +7,6 @@
     "fullName": {
       "type": "object",
       "properties": {
-        "salutation": {
-          "type": "string"
-        },
         "first": {
           "type": "string",
           "minLength": 1,
@@ -24,6 +21,7 @@
           "maxLength": 30
         },
         "suffix": {
+          "type": "string",
           "enum": [
             "Jr.",
             "Sr.",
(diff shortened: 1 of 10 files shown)


dist/edu-benefits-schema.json
```diff
@@ -216,9 +216,6 @@
     "fullName": {
       "type": "object",
       "properties": {
-        "salutation": {
-          "type": "string"
-        },
```
lihanli: is this not used in the 1990?

jbalboni: No, our name component used in hca and the 1990 doesn't have this field.

src/change-of-program/schema.js
```diff
@@ -18,7 +18,8 @@ export default {
     'serviceBefore1977',
     'date',
     'dateRange',
-    'educationType'
+    'educationType',
+    'preferredContactMethod',
```
lihanli: trailing comma

dist/change-of-program-schema.json
```diff
@@ -338,6 +326,14 @@
...
         "cooperativeTraining"
       ]
+    },
+    "preferredContactMethod": {
+      "type": "string",
+      "enum": [
+        "mail",
+        "email",
+        "phone"
+      ]
```
lihanli: can you add this definition to the 1990 schema?

jbalboni: Done.

PR #57, opened 2017-02-13
1990e schema

lihanli:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1078


dist/edu-benefits-schema.json
```diff
@@ -326,7 +326,8 @@
         "flightTraining",
         "testReimbursement",
         "licensingReimbursement",
-        "tuitionTopUp"
+        "tuitionTopUp",
+        "cooperativeTraining"
```
jbalboni: I don't think we want to allow this option in this schema?

src/schemas/transfer-benefits/schema.js
```diff
@@ -0,0 +1,105 @@
...
+    'bankAccount',
+    'educationType',
+    'school',
+    'postHighSchoolTrainings',
+    'dateRange',
+    'nonMilitaryJobs'
+  ),
+  properties: {
+    privacyAgreementAccepted: {
+      $ref: '#/definitions/privacyAgreementAccepted'
```
jbalboni: I'm starting to dislike `$ref`, since it means we have two different ways of sharing definitions. It doesn't seem like it buys us much, given that we're already importing shared definitions and could easily just set fields to the object directly:

```
privacyAgreementAccepted: definitions.privacyAgreementAccepted
```

`$ref` does make the schema size a bit smaller, but outside of the address definitions (which has countries and states listed), it doesn't seem like much of a difference.

I'm not saying we should change anything right now, more just curious about your thoughts @lihanli @markolson.

dist/definitions.json
```diff
@@ -335,5 +335,133 @@
     "enum": [
       true
     ]
+  },
+  "gender": {
+    "type": "string",
+    "enum": [
+      "M",
```
jbalboni: Can you swap the order of these? We've been ordering this Female, Male on the front end, so I'll end up changing it when we build the form anyway.

package.json
```diff
@@ -14,6 +14,9 @@
   "files": [
     "dist"
   ],
+  "dependencies": {
+    "coding-inflector": "0.0.1"
```
jbalboni: This seems handy.

src/schemas/transfer-benefits/schema.js
lihanli: i think we should keep definitions, the address definition is used in a lot of places (school.address, veteranAddress, spouseAddress, and relativeAddress) plus some definitions use other definitions

PR #58, opened 2017-02-16
Change direct deposit options

jbalboni:
We're going to combine the start/update options into one, according to [this issue](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1163).


jbalboni:
@markolson Added another option here.

PR #60, opened 2017-02-24
Change order of direct deposit change options

jbalboni:
Related to [1354](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1354)


lihanli:
need to update version

PR #67, opened 2017-03-09
Add education start date to 1990e

jbalboni:
The outline in [this issue](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/642) has a start date, so I'm adding that.


lihanli:
i talked to alex about it since that field isn't in the paper form, he says he's going to check with VA people if we need it

PR #68, opened 2017-03-10
Form 1990n

lihanli:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1593


src/common/definitions.js
```diff
@@ -311,6 +311,21 @@ const educationProgram = {
...
+const currentlyActiveDuty = {
+  type: 'object',
+  properties: {
+    yes: {
+      type: 'boolean'
+    },
+    onTerminalLeave: {
+      type: 'boolean'
+    },
+    nonVaAssistance: {
```
jbalboni: Why is this field in here? They don't seem directly related (and they're on different pages in the UI).

src/schemas/ncs-benefits/schema.js
```diff
@@ -0,0 +1,49 @@
...
+let schema = {
+  $schema: 'http://json-schema.org/draft-04/schema#',
+  title: "APPLICATION FOR VA EDUCATION BENEFITS UNDER THE NATIONAL CALL TO SERVICE (NCS) PROGRAM (22-1990N)",
+  type: 'object',
+  additionalProperties: false,
+  definitions: _.pick(definitions, [
+    'educationType',
+    'dateRange'
+  ]),
+  properties: {
```
jbalboni: There's a question on the GH [issue](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/632), "What is the source of these funds?" that I don't see.

Also, "Are you getting, or do you expect to get any money (including, but not limited to, federal tuition assistance) from the Armed Forces or public health services for any part of your coursework or training?" from [here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/633).

And preferredContactMethod ([here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/634))

src/common/definitions.js
lihanli: this is supposed to be
```
11B. FOR ACTIVE DUTY CLAIMANTS ONLY: Are you receiving or do you anticipate receiving
 any money (including but not limited to Federal Tuition Assistance) from the Armed Forces
 or Public Health Service for the course for which you have applied to the VA for education
 benefits? If you receive such benefits during any part of your training, check "YES."
```
it's in the same object because the question is for people on active duty only. also it was in the same object for the 1990 and i wanted to keep things organized the same way whenever possible.

jbalboni: Ok. I'll check with Alex and see if the UI needs to change, since I don't think we're limiting that field to active duty vets only.

PR #71, opened 2017-03-16
1990e add va file number

annekainicUSDS:
Adding VA file number to the 1990e from feedback we received during testing with Shay: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1657. Add it to the schema so I can make the change on the front-end.


lihanli:
can you update the version in package.json?

src/schemas/transfer-benefits/schema.js
```diff
@@ -8,7 +8,7 @@ let schema = {
   type: 'object',
   additionalProperties: false,
   definitions: _.pick(definitions,
-    'privacyAgreementAccepted',
+    'privacyAgreementAccepted', [
```
lihanli: not needed i think?

annekainicUSDS: You're right, I don't think they're necessary, I was just copying from change-of-program, which has them. I can remove.

annekainicUSDS:
@lihanli updates made!

jbalboni:
FYI @raquelromano-gov, I think you have an item related to this.

dist/transfer-benefits-schema.json
```diff
@@ -458,8 +458,24 @@
...
   },
+  "anyOf": [
+    {
+      "required": [
+        "vaFileNumber"
+      ]
+    },
+    {
+      "required": [
+        "relativeSocialSecurityNumber"
```
raquelromano-gov: @annekainicUSDS @lihanli 
Just noticed this--shouldn't this be veteranSocialSecurityNumber?

I believe the relative is the applicant (relative of the veteran), but the veteran is the one who can offer a va file # in lieu of a social security number.

PR #72, opened 2017-03-17
Modified toursOfDuty for 5490

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1532

Removes `benefitsToApplyTo` from the 5490. This also means it doesn't use a definition, but I didn't want to remove it from the definition in case it's used somewhere else. If not, that can be addressed later.

### DIFF
--- a/src/schemas/dependents-benefits/schema.js
+++ b/src/schemas/dependents-benefits/schema.js
@@ -4,6 +4,8 @@ import _ from 'lodash';
 
 let definitions = _.cloneDeep(originalDefinitions);
 definitions.educationType.enum.push('farmCoop');
+const modifiedToursOfDuty = definitions.toursOfDuty;
+delete modifiedToursOfDuty.items.properties.benefitsToApplyTo;
 
 let schema = {
   $schema: 'http://json-schema.org/draft-04/schema#',
@@ -99,6 +101,7 @@ let schema = {
         }
       )
     },
+    toursOfDuty: modifiedToursOfDuty,
     civilianBenefitsAssistance: {
       type: 'boolean'
     },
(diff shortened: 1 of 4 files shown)


cvalarida:
I'm not really sure how to modify the unit test to get it up and running; I'll look into it next week, but if somebody could point me in the right direction to help me along, that'd be much appreciated.

jbalboni:
@cvalarida In test/dependents-benefits/schema.spec.js, I think you need to remove toursOfDuty from the array on line 24 and add tests for your modified version in that file instead.

src/schemas/dependents-benefits/schema.js
```diff
@@ -137,7 +140,7 @@ let schema = {
   ['date', 'benefitsRelinquishedDate'],
   ['fullName', 'previousBenefits.veteranFullName'],
   ['ssn', 'previousBenefits.veteranSocialSecurityNumber'],
-  ['toursOfDuty'],
+  // ['toursOfDuty'],
```
lihanli: do we need this comment here?

lihanli:
you can add a test like this
```
schemaTestHelper.testValidAndInvalid('toursOfDuty', {
  valid: [valid data],
  invalid: [invalid data]
});
```

src/schemas/dependents-benefits/schema.js
cvalarida: No, I thought I deleted that before I committed. It'll be gone.

test/schemas/dependents-benefits/schema.spec.js
```diff
@@ -80,6 +79,18 @@ describe('dependents benefits schema', () => {
     }]
   });
 
+  schemaTestHelper.testValidAndInvalid('toursOfDuty', {
+    valid: [[{
```
cvalarida: Took me entirely too long to figure out an array with an object _inside_ the `valid` array. 0.o

cvalarida:
@lihanli Does anything need to happen for the back end to make this work, or will it be good to go once this is merged and updated in vets-api package.json?

lihanli:
i can update the backend after this is merged

PR #73, opened 2017-03-20
allow partial addresses

lihanli:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1502

### DIFF
--- a/test/common/definitions.spec.js
+++ b/test/common/definitions.spec.js
@@ -37,13 +37,13 @@ describe('schema definitions', () => {
         city: 'abc',
         state: 'VA',
         country: 'USA'
-      }
-    ],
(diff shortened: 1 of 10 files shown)


test/common/definitions.spec.js
```diff
@@ -37,13 +37,13 @@ describe('schema definitions', () => {
...
-    ],
-    invalid: [
+      },
       {
         city: 'foo',
         country: 'USA'
-      },
+      }
+    ],
+    invalid: [
```
cvalarida: Is this only invalid because the state is `'foo'`?

lihanli: yea, the state value has to match the enum

PR #76, opened 2017-03-24
Move SSN/FN choice to sponsor/veteran

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1744 (although there was never an explicit backend ticket for this)

### DIFF
--- a/dist/transfer-benefits-schema.json
+++ b/dist/transfer-benefits-schema.json
@@ -467,7 +467,7 @@
     },
     {
       "required": [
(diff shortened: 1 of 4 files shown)


annekainicUSDS:
Actually one thing I forgot, we might want to update the version number now. Like Lihan asked me to do here: https://github.com/department-of-veterans-affairs/vets-json-schema/pull/71

raquelromano-gov:
OK, updated version number. Will merge now--thanks!

PR #80, opened 2017-04-05
Add graduation expected date

annekainicUSDS:
Added a graduation expected date. Not sure if I updated the package version correctly, so let me know :)

### DIFF
--- a/dist/22-5490-schema.json
+++ b/dist/22-5490-schema.json
@@ -679,6 +679,9 @@
         },
         "highSchoolOrGedCompletionDate": {
           "$ref": "#/definitions/date"
+        },
+        "graduationExpectedDate": {
+          "$ref": "#/definitions/date"
         }
       }
     },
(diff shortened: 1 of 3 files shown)


lihanli:
highSchoolOrGedCompletionDate is supposed to be the graduation expected date, the name is unclear because the field name is copied from the 1990

markolson:
Using `highSchoolOrGedCompletionDate ` would fill in this date in the "Applicant's date of disposition: " field of the spool file. Did we want different language around that, @annekainicUSDS?

annekainicUSDS:
Ah okay, I was thinking it was only used for a past graduation date. But you're right I could use it for a future date too. Just need to change the front-end validations based on the situation. So a schema change might not be needed then?

markolson:
I think that's right, just sent it over as the `highSchoolOrGedCompletionDate` and we're good!

annekainicUSDS:
Perfect! Will close this PR then.

PR #81, opened 2017-04-07
allow either SSN or VA FN for sponsor veteran

lihanli:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1929

### DIFF
--- a/src/schemas/22-5490/schema.js
+++ b/src/schemas/22-5490/schema.js
@@ -147,6 +147,7 @@ let schema = {
   ['date', 'spouseInfo.remarriageDate'],
   ['date', 'benefitsRelinquishedDate'],
   ['fullName', 'previousBenefits.veteranFullName'],
+  ['vaFileNumber', 'previousBenefits.vaFileNumber'],
   ['ssn', 'previousBenefits.veteranSocialSecurityNumber'],
   ['date', 'highSchool.highSchoolOrGedCompletionDate'],
   ['postHighSchoolTrainings'],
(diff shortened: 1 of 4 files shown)


src/schemas/22-5490/schema.js
```diff
@@ -68,7 +70,7 @@ let schema = {
       type: 'string',
       enum: ['chapter35', 'chapter33']
     },
-    previousBenefits: {
+    previousBenefits: Object.assign({
```
raquelromano-gov: Any reason to use `Object.assign()` instead of `_.assign()` here and elsewhere?

test/schemas/22-5490/schema.spec.js
```diff
@@ -14,7 +14,11 @@ const schemaDefaults = {
   }
 };
 
-let schemaTestHelper = new SchemaTestHelper(_.omit(schema, 'anyOf'), schemaDefaults);
+let liteSchema = _.cloneDeep(schema);
+delete liteSchema.anyOf;
```
raquelromano-gov: I only vaguely understand why we need to treat `anyOf` fields in our schemas differently. Can you add a comment here and below where you call `_.omit(schema, 'anyOf')` explaining the reason?

jbalboni: Yeah, I'm not sure I understand why it's being removed here.

src/schemas/22-5490/schema.js
lihanli: if there's a file that hasn't included lodash i can use `Object.assign` without including anything, and if i'm going to do that then i should use it everywhere for consistency

PR #85, opened 2017-04-13
Remove anyOf from 5490 root

jbalboni:
We don't force people to enter a file number of the 5490, so remove that check.

### DIFF
--- a/test/schemas/22-5490/schema.spec.js
+++ b/test/schemas/22-5490/schema.spec.js
@@ -138,6 +138,4 @@ describe('dependents benefits schema', () => {
       name: 1
     }]
   });
-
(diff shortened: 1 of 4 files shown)


test/schemas/22-5490/schema.spec.js
```diff
@@ -139,5 +139,5 @@ describe('dependents benefits schema', () => {
     }]
   });
 
-  (new SharedTests(new SchemaTestHelper(schema, schemaDefaults))).requireSsnOrFile();
+  new SharedTests(new SchemaTestHelper(schema, schemaDefaults));
```
lihanli: you can delete this line, it is for the anyOf test

jbalboni: Removed

PR #86, opened 2017-04-17
WIP: Switch farmCoop to cooperativeTraining for 5490

annekainicUSDS:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2043

All forms that used `farmCoop` should now be using `cooperativeTraining` instead. This currently only applied to the 5490.

### DIFF
--- a/dist/22-5490-schema.json
+++ b/dist/22-5490-schema.json
@@ -25,7 +25,7 @@
         "testReimbursement",
         "licensingReimbursement",
         "tuitionTopUp",
-        "farmCoop"
+        "cooperativeTraining"
       ]
     },
     "privacyAgreementAccepted": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "2.5.2",
+  "version": "2.5.3",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/department-of-veterans-affairs/vets-json-schema.git"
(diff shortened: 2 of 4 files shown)


dist/22-5490-schema.json
```diff
@@ -25,7 +25,7 @@
         "testReimbursement",
         "licensingReimbursement",
         "tuitionTopUp",
-        "farmCoop"
+        "cooperativeTraining"
```
annekainicUSDS: Question for @lihanli or @markolson: Now that we're switching the options for this field from `farmCoop` to `cooperativeTraining`, what is someone is in the middle of completing this form while we deploy this change, had selected `farmCoop` and then submit the form, their submission won't pass the new schema, correct? How should we account for this?

lihanli: we should add cooperativeTraining and leave farmCoop in the backend/schema for a week to avoid errors, deploy the changes to frontend first and then remove farmCoop from the backend schema later

markolson: We'll have to remember to do that in the spool file as well.

annekainicUSDS: There actually wasn't a front-end change for this, as the front-end is automatically populating the dropdown options from the schema. It sounds like I'll have to make a corresponding front-end change that drops `farmCoop` from the list, and add it back here in the schema for a week, like @lihanli suggested. Tackling that now!

annekainicUSDS:
Not sure what the outcome of this was, last I heard we were waiting on the education services team for final confirmation. Going to close for now and it can be reopened if necessary.

PR #87, opened 2017-04-20
Update hca schema for new rjsf version

annekainicUSDS:
Updates needed for new rjsf version of HCA:
1. Specify `type` for a few of the fields (rjsf won't render the fields without it).
2. Remove non-US states from the possible options for `stateOfBirth`. For the current version of HCA, the front-end filters these out. Since the rjsf version of HCA will populate the options directly from the schema, it seemed to make sense to just remove them here, as they aren't needed anyway. I also tried using `updateSchema` in `config.js`, but that didn't work for me. This seems like a better option, anyway.

Real question is: can this be merged now? It will affect the current version of HCA in production. I don't think anything would negatively impact that, but a review on this would be great!


dist/10-10EZ-schema.json
```diff
@@ -361,52 +362,8 @@
       "maxLength": 20
     },
     "stateOfBirth": {
+      "type": "string",
```
jbalboni: Should we go through and fix all of these? I found one (vaMedicalFacility) for the pages I'm working on. It's consistently on fields with `enum`.

annekainicUSDS: Just added those in all the places that appeared to be missing a `type`

annekainicUSDS:
Undid the state changes because I found a way to handle this on the front-end.

PR #88, opened 2017-04-21
Remove top up from education type options on 1990n

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2317

We don't want the tuitionTopUp option in the ed types dropdown on the 1990n.


lihanli:
version change

PR #89, opened 2017-04-28
Remove top up option from 5495

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2494


jbalboni:
@J-Quag There are two parts to this change and it looks like I had forgotten to put up a PR for the other one. It's up now.

PR #91, opened 2017-05-10
Add relative vaFileNumber

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2733

Open to other suggestions about how to disambiguate the two file numbers.


markolson:
I'm ok with this - it sort of fits in with how we do SSN when there's a relative and veteran. @lihanli ?

PR #94, opened 2017-05-23
Reorganized financials and dependents

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2911

The EZ form has somewhat simplified financials (e.g. no interest) and the current UI for the financials section has us entering financial data for each spouse and child. It's very difficult to accomplish this with the current schema, so I've moved the financial info to be part of each row in the children array. I've also consolidated that array and made the in/out of household separation a property in the schema, instead of separate arrays.

I haven't updated tests or the version, I wanted to get some feedback first.

### DIFF
--- a/src/schemas/21-527/schema.js
+++ b/src/schemas/21-527/schema.js
@@ -7,10 +7,65 @@ let schema = {
   title: 'INCOME, NET WORTH, AND EMPLOYMENT STATEMENT',
   type: 'object',
   additionalProperties: false,
-  definitions: _.pick(definitions,
-    'dateRange',
-    'netWorthAccount'
-  ),
+  definitions: _.merge(_.pick(definitions,
+    'dateRange'
+  ), {
+    netWorth: {
+      type: 'object',
+      properties: {
+        bank: { type: 'integer' },
+        ira: { type: 'integer' },
+        stocks: { type: 'integer' },
+        business: { type: 'integer' },
+        realProperty: { type: 'integer' },
+        otherProperty: { type: 'integer' },
+        additionalSources: { $ref: '#/definitions/additionalSources' }
+      }
+    },
+    additionalSources: {
+      type: 'array',
+      items: {
+        type: 'object',
+        properties: {
+          name: {
+            type: 'string'
+          },
+          amount: {
+            type: 'integer'
+          }
+        }
+      }
+    },
+    monthlyIncome: {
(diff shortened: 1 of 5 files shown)


src/schemas/21-527/schema.js
```diff
@@ -7,10 +7,80 @@ let schema = {
...
-    'dateRange',
-    'netWorthAccount'
-  ),
+  definitions: _.merge(_.pick(definitions,
+    'dateRange'
+  ), {
+    netWorth: {
+      type: 'object',
+      properties: {
+        bank: { type: 'integer' },
```
lihanli: in the form this is separated into interest bearing and non interest bearing bank accounts

src/schemas/21-527/schema.js
```diff
@@ -7,10 +7,80 @@ let schema = {
...
+          amount: {
+            type: 'integer'
+          }
+        }
+      }
+    },
+    monthlyIncome: {
+      type: 'object',
+      properties: {
+        salary: {
```
lihanli: 527EZ doesn't have salary but adds a field called "service retirement" which is separate from civil service

src/schemas/21-527/schema.js
jbalboni: In the full 527, yes, but in the 527EZ, it's just the amount.

lihanli: in this 527EZ which looks like the latest has them separated https://www.vba.va.gov/pubs/forms/VBA-21-527EZ-ARE.pdf

jbalboni: Yep, I misread your comment, sorry.

src/schemas/21-527/schema.js
```diff
@@ -7,10 +7,77 @@ let schema = {
...
+    expectedIncome: {
+      type: 'object',
+      properties: {
+        salary: {
+          type: 'integer'
+        },
+        interest: {
+          type: 'integer'
+        },
+        other: {
```
lihanli: this isn't needed when additionalSources is added

jbalboni: I agree, but I'm confirming that with Alex.

jbalboni: @lihanli Alex wants to keep it for now until we can run it by stakeholders.

PR #100, opened 2017-06-01
Add activation date as 'date' to nationalGuard

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#2222

Adds date to 527EX nationalGuard schema so that can match design as:

Are you currently on federal active duty in the National Guard? * [Y/N]
{ if yes }
Date of activation [DATE]
Name of Reserve/NG unit [TEXT]
Address of unit [ADDRESS]
Phone number of unit [PHONE]

### DIFF
--- a/dist/21P-527EZ-schema.json
+++ b/dist/21P-527EZ-schema.json
@@ -472,6 +472,9 @@
         },
         "phone": {
           "$ref": "#/definitions/usaPhone"
+        },
+        "date": {
+          "$ref": "#/definitions/date"
         }
       }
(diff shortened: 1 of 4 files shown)


lihanli:
can you delete the activationDate field and update the version in package.json?

cehsu:
Thanks @lihanli, changes made.

lihanli:
LGTM once the tests are fixed, replace `activationDate` with `nationalGuard.date` in the tests, it will test the right field with that syntax

PR #101, opened 2017-06-02
Update 527EZ schema to limit severance pay to amount only

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#2223


lihanli:
severance pay type is in the form listed under question 16C, maybe UX missed it?

cehsu:
Thanks @lihanli, I'll close this, unless it turns out that this change makes sense.

PR #105, opened 2017-06-13
Remove other field from expected income

lihanli:
expected_income
there's no place to put an unnamed "other" expected income, the other fields all need names which are included in additional sources


jbalboni:
@lihanli I think @goldenmeanie said we didn't need the separate other field, additional sources was enough.

PR #106, opened 2017-06-13
Provide claimant contact information fields

cehsu:
Includes address, email, and phone.


lihanli:
can you add the changes to `src/schemas/21P-530/schema.js` and add specs?

cehsu:
Thanks @lihanli, changes made.

cehsu:
Thanks @lihanli, changes made but causing errors: 
```
1) 21-530 schema should allow address with {"street":"123 a rd","city":"abc","country":"USA"}:

      AssertionError: expected false to equal true
      + expected - actual

      -false
      +true
      
      at SchemaTestHelper.schemaExpect (test/support/schema-test-helper.js:34:42)
      at Context.<anonymous> (test/support/schema-test-helper.js:47:16)

  2) 21-530 schema shouldnt allow address with {"city":1}:
     AssertionError: expected '' to include '.address'
      at SchemaTestHelper.schemaExpect (test/support/schema-test-helper.js:37:46)
      at Context.<anonymous> (test/support/schema-test-helper.js:47:16)

  3) 21-530 schema should allow phone with "555-555-5555":

      AssertionError: expected false to equal true
      + expected - actual

      -false
      +true
      
      at SchemaTestHelper.schemaExpect (test/support/schema-test-helper.js:34:42)
      at Context.<anonymous> (test/support/schema-test-helper.js:47:16)

  4) 21-530 schema should allow phone with "555 555 5555":

      AssertionError: expected false to equal true
      + expected - actual

      -false
      +true
      
      at SchemaTestHelper.schemaExpect (test/support/schema-test-helper.js:34:42)
      at Context.<anonymous> (test/support/schema-test-helper.js:47:16)

```
Any initial thoughts?

lihanli:
is address supposed to be the same as the one in `src/common/definitions.js`?

cehsu:
Good question, @lihanli. That definition includes all countries, I assumed I should follow the example in 527EZ (USA/CAN/MEX only). Is there a good place to determine which it should be?

lihanli:
in the form i don't see anything about the address being USA/CAN/MEX only, are you sure that's a requirement?

lihanli:
add `['address', 'veteranAddress'],` to this array and that will import it from definitions.js and add it to the schema. we should probably name it veteranAddress to keep the naming consistent with the other schemas if the address is referring to the veterans address.

```js
[
  ['fullName', 'claimantFullName'],
  ['fullName', 'veteranFullName'],
  ['ssn', 'veteranSocialSecurityNumber'],
  ['vaFileNumber'],
  ['date', 'burialDate'],
  ['date', 'deathDate'],
].forEach((args) => {
  schemaHelpers.addDefinitionToSchema(schema, ...args);
});
```

cehsu:
Thanks @lihanli, good point. I have made the change, except I have stuck with `address` as that is the name outlined in the [relevant issue](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3152), and as the address refers to the claimant.

lihanli:
for the address test you need to add `sharedTests.runTest('address', ['address']);`, right now it's testing the default field name `veteranAddress` which doesn't exist.

do you know if we are filling out a PDF for this form? if so we still need usaPhone instead of the phone definition from definitions.js

cehsu:
Thanks @lihanli, I've made clearer labels and used the usaPhone format. I removed the test for the latter as that was also done for the 527EZ, please let me know if that should be replaced.

src/schemas/21P-530/schema.js
```diff
@@ -40,17 +44,25 @@ let schema = {
...
+    claimantEmail: {
+      type: 'string',
+      format: 'email'
+    },
+    claimantPhone: schemaHelpers.getDefinition('usaPhone'),
   }
 };
 
 [
+  ['address'],
```
lihanli: can you change this to `['address', 'claimantAddress']` and remove claimantAddress above?

src/schemas/21P-530/schema.js
```diff
@@ -40,17 +44,25 @@ let schema = {
...
   ['fullName', 'claimantFullName'],
   ['fullName', 'veteranFullName'],
   ['ssn', 'veteranSocialSecurityNumber'],
   ['vaFileNumber'],
   ['date', 'burialDate'],
   ['date', 'deathDate'],
+  ['usaPhone'],
```
lihanli: this isn't needed if it's already been added manually. what `schemaHelpers.addDefinitionToSchema` does is try to import the definition from definitions.js if it doesn't already exist. then first argument is the definition name and the 2nd argument is the name of the field to put the definition in

lihanli:
you can just leave usaPhone like that for now i can add a definition and a shared test for both forms later

PR #109, opened 2017-06-15
Add benefit fields to burials schema

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3148

### DIFF
--- a/src/schemas/21P-530/schema.js
+++ b/src/schemas/21P-530/schema.js
@@ -78,53 +78,53 @@ let schema = {
       format: 'email'
     },
     claimantPhone: schemaHelpers.getDefinition('usaPhone'),
-		burialAllowance: {
-			type: 'boolean'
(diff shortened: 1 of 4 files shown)


src/schemas/21P-530/schema.js
```diff
@@ -78,6 +78,53 @@ let schema = {
...
+      'enum': [
+        'nonService',
+        'service',
+        'vaMC'
+      ]
+    },
+    burialCost: {
+      type: 'number'
+    },
+    previouslyReceiveAllowance: {
```
cehsu: Should this be `previouslyReceivedAllowance`?

jbalboni: Yeah, that's probably better

jbalboni:
@cehsu Updated

PR #110, opened 2017-06-15
530 claimant contact confirmation email

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#3152
Based on feedback to this [PR](https://github.com/department-of-veterans-affairs/vets-website/pull/5752#issuecomment-308832233).


src/schemas/21P-530/schema.js
```diff
@@ -124,7 +119,16 @@ let schema = {
     },
     amountGovtContribution: {
       type: 'number'
-    }
+    },
+    claimantEmail: {
+      type: 'string',
+      format: 'email'
+    },
+    'view:claimantEmailConfirmation': {
```
lihanli: aren't these fields not supposed to be in the schema?

cehsu: Thanks @lihanli, @J-Quag and @stacyhlinn provided feedback on [this PR](https://github.com/department-of-veterans-affairs/vets-website/pull/5752#issuecomment-308832233), indicating that the email should be checked as in EDU. Please let me know if I should approach this differently.

lihanli: i think the view fields are supposed to be used only in on the frontend and removed before sending to the backend, you can add this field after importing the schema and do the email check entirely in the frontend

cehsu: Thanks for the reminder @lihanli, I'll close this.

PR #112, opened 2017-06-16
Updates for dependents on pensions

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2905

Parent and child dependents will be in the this same array, since they both have the same financial info and are entered together in the UI. Spouses are still separate (the above ticket will be updated).

Most fields are still for children only, so I left the `child` prefixes except for the name.

### DIFF
--- a/dist/21P-527EZ-schema.json
+++ b/dist/21P-527EZ-schema.json
@@ -580,13 +580,20 @@
         }
       }
     },
-    "children": {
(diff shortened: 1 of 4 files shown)


jbalboni:
@lihanli It's implied in the financials section: "...net worth of your dependents (spouse, child, etc.),”

PR #114, opened 2017-06-19
Add files to burials and pensions schemas

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3326

This adds a common field for documents to the schema. The data needed for the backend is the confirmationCode field, which will have the identifier sent back from the upload endpoint. The other fields are probably not useful, but this is less error-prone that me stripping them out.

It's also easier on the front end to have this always be an array, rather than be an object or array depending on if multiple files are allowed. There's validation on the front end to set the min and max items. We could do that here too if it seems useful.

### DIFF
--- a/test/common/definitions.spec.js
+++ b/test/common/definitions.spec.js
@@ -255,4 +255,15 @@ describe('schema definitions', () => {
   ].forEach((definition) => {
     testValidAndInvalidDefinitions(definition, testData[definition].data);
   });
+
+  testValidAndInvalidDefinitions('files', {
+    valid: [[{
+      confirmationCode: 'testing',
+      name: 'testing',
+      size: 1
+    }]],
+    invalid: [[{
+      size: 'asdf'
+    }]]
+  });
 });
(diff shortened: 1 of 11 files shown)


test/common/definitions.spec.js
```diff
@@ -255,4 +255,15 @@ describe('schema definitions', () => {
...
+  testValidAndInvalidDefinitions('files', {
+    valid: [[{
+      confirmationCode: 'testing',
+      name: 'testing',
+      size: 1
+    }]],
+    invalid: [[{
+      size: 'asdf'
+    }]]
+  });
```
lihanli: can you delete this and add `files` to this block?
```js
  [
    'toursOfDuty',
    'currentlyActiveDuty',
    'bankAccountChange',
    'usaPhone',
    'maritalStatus',
    'relationshipAndChildName',
    'marriages'
  ].forEach((definition) => {
    testValidAndInvalidDefinitions(definition, testData[definition].data);
  });
```

jbalboni:
@lihanli Done.

PR #118, opened 2017-06-21
527 ez dependent information

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#2907

As personWhoLivesWithChild is never required (not even when the applicant specifies that their dependent child does not live with them), the required properties of the default fullName definition must be omitted.

Pointed out by @jbalboni [here](https://github.com/department-of-veterans-affairs/vets-website/pull/5802#pullrequestreview-45315319).


lihanli:
personWhoLivesWithChild is already not required, your test passes on master without any schema changes

PR #121, opened 2017-06-23
Update expenses field

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3484

The UI for other expenses is presented like the financials fields, so this changes the schema to match those.

### DIFF
--- a/dist/21P-527EZ-schema.json
+++ b/dist/21P-527EZ-schema.json
@@ -37,6 +37,10 @@
         }
       }
     },
+    "date": {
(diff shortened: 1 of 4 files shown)


lihanli:
you're allowed to put dependent expenses in other expenses so this will need to be added to the dependents array, seems more complicated than what we had originally but if it works better for the frontend i can work around it

jbalboni:
@lihanli I did add `otherExpenses` to the `dependents` array.

Yeah, this is more complicated. Alex wanted it to follow the pattern for the other financial data, so I had to lay it out like those fields. If it's a huge pain, though, we can check with him, though.

PR #123, opened 2017-06-23
Group PDRL and TDRL

cvalarida:
Connected to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3485

Without testing it (didn't want to mess with `npm link` right now), I _assume_ this moves the order.


jbalboni:
@cvalarida It does move the order.

PR #126, opened 2017-06-28
burials address change

lihanli:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3586
old address format doesn't work with burials form

### DIFF
--- a/dist/21P-530-schema.json
+++ b/dist/21P-530-schema.json
@@ -4,6 +4,42 @@
   "type": "object",
   "additionalProperties": false,
   "definitions": {
+    "address": {
+      "type": "object",
+      "properties": {
+        "street": {
+          "type": "string"
+        },
+        "aptNum": {
+          "type": "string",
+          "minLength": 1,
+          "maxLength": 5
+        },
+        "city": {
+          "type": "string"
+        },
+        "state": {
+          "type": "string",
+          "minLength": 2,
+          "maxLength": 2
+        },
+        "country": {
+          "type": "string",
+          "minLength": 2,
+          "maxLength": 2
+        },
+        "postalCode1": {
+          "type": "string",
+          "minLength": 5,
+          "maxLength": 5
+        },
+        "postalCode2": {
+          "type": "string",
+          "minLength": 4,
+          "maxLength": 4
+        }
+      }
+    },
     "dateRange": {
       "type": "object",
       "properties": {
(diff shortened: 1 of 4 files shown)


dist/21P-530-schema.json
```diff
@@ -4,6 +4,42 @@
...
+        },
+        "state": {
+          "type": "string",
+          "minLength": 2,
+          "maxLength": 2
+        },
+        "country": {
+          "type": "string",
+          "minLength": 2,
+          "maxLength": 2
```
cehsu: not 3?

lihanli: only room for 2 characters on the form

cehsu: thanks for the explanation @lihanli

dist/21P-530-schema.json
```diff
@@ -4,6 +4,42 @@
...
+        "street": {
+          "type": "string"
+        },
+        "aptNum": {
+          "type": "string",
+          "minLength": 1,
+          "maxLength": 5
+        },
+        "city": {
+          "type": "string"
```
cehsu: No length restrictions here? It looks like 18 on the form.

dist/21P-530-schema.json
```diff
@@ -4,6 +4,42 @@
   "type": "object",
   "additionalProperties": false,
   "definitions": {
+    "address": {
+      "type": "object",
+      "properties": {
+        "street": {
+          "type": "string"
```
cehsu: Should this have a max length of 30?

dist/21P-530-schema.json
lihanli: for the longer fields i have overflow code that will put "see additional information page" in the field and add an extra page with the answer if it goes over the limit, technically we could use this for all the fields but there's no space to put the "see last page" text for fields that have <5 characters available so i would have to make a custom solution for this section where all the address fields are moved to the additional info page and it would be a lot more complicated for me, we could do it if it was really important to  support addresses that don't fit the form though

src/schemas/21P-530/schema.js
```diff
@@ -34,6 +34,42 @@ let schema = {
   type: 'object',
   additionalProperties: false,
   definitions: {
+    address: {
```
jbalboni: This seems like it's going to have implications for prefilling this form, @markolson?

src/schemas/21P-530/schema.js
```diff
@@ -34,6 +34,42 @@ let schema = {
...
+          type: 'string',
+          minLength: 2,
+          maxLength: 2
+        },
+        country: {
+          type: 'string',
+          minLength: 2,
+          maxLength: 2
+        },
+        postalCode1: {
```
jbalboni: This seems limiting for something like a Canadian postal code, which is six characters, or other countries that might have different formats.

src/schemas/21P-530/schema.js
```diff
@@ -34,6 +34,42 @@ let schema = {
...
+    address: {
+      type: 'object',
+      properties: {
+        street: {
+          type: 'string'
+        },
+        aptNum: {
+          type: 'string',
+          minLength: 1,
+          maxLength: 5
```
jbalboni: A lot of the validation in here is really restrictive, and I realize that's because of the form boxes, but when a user is normally filling out the form they can write outside the boxes or put their address on another page. They don't have that option in our form with these kinds of limits.

markolson:
If the primary issue with some of these restrictions are around the `see additional...` text, could we have a convention of `X`ing each character in the field out, for instance? Like Jeff mentioned, this will impact international claimants as well as prefill data, so I'd _prefer_ that we could work around it somehow, but this is a good conversation to have.

lihanli:
if we can always put the address on the additional info page that would work for me, it would be way easier to put the entire form as additional info and skip filling the pdf

lihanli:
@J-Quag @goldenmeanie we need some help on how do address input for burials. here are the options

1. make address input as restrictive as it is on the paper form (only 2 letters allowed for state/province, only US postal codes allowed). easy to implement.
2. mark overflow fields with `XXX` instead of the current text `See add'l info page` and make sure whoever is in charge of processing forms knows about this. easy to implement.
3. when any field in the address has an overflow put `See add'l info page` in the `No. & Street` field, leave all the other ones blank, and move the entire address to the additional info page. harder to implement.
4. same as option 3 but always put the address on the additional info page regardless of whether or not there is an overflow. easy to implement.

goldenmeanie:
Is option 2 about having a string short enough to fit in any field?

I am in favor of moving logical chunks to the overflow page. Option 3 seems like the nicer experience, but 4 seems like a decent fallback.

Do we have any sense of the average number of overflowed fields per application?

jbalboni:
Unless we make changes to our address field schema, option 3 will always result in an overflow address, since we use 3 digit country codes and the form only has 2.

lihanli:
yea, option 2 is about the overflow text being too big to fit in a small field.

for the all the options we should still change the schema to replace street2 with apt number and country code to be 2 letters

PR #136, opened 2017-07-31
Make all other expenses fields required

cehsu:
Do we still need disabilityOrRelationship: 'disability' in the spec for this?


cehsu:
@lihanli thanks for your swift review, do we still need disabilityOrRelationship: 'disability' in the spec for this?

lihanli:
that can be deleted it used to be a field in the schema but was deleted earlier

PR #143, opened 2017-08-15
Create seniorRotcCommissioned and make seniorRotc fields not required

cehsu:
These changes provide another boolean property `seniorRotcCommissioned`, and make seniorRotc fields not required, as it looks like they're not required on the current form. Please let me know if we decided to update this, or if it makes sense to, and I'll revert that part of the changes.

screen shot 2017-08-15 at 1 57 00 pm

### DIFF
--- a/dist/22-1990-schema.json
+++ b/dist/22-1990-schema.json
@@ -627,10 +627,10 @@
             }
           }
         }
-      },
-      "required": [
-        "rotcScholarshipAmounts"
-      ]
+      }
+    },
+    "seniorRotcCommissioned": {
+      "type": "boolean"
     },
     "seniorRotcScholarshipProgram": {
       "type": "boolean"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "3.0.34",
+  "version": "3.0.35",
   "repository": {
     "type": "git",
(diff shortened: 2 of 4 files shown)


jbalboni:
@cehsu We don't need to add `seniorRotcCommissioned`, we can treat it as a `view:` field on the front end. The backend doesn't use it.

Also, this schema is currently used by the backend, so the current form is submitting data that validates with it. It passes with the `required` value you changed because we don't send the `seniorROTC` object if there are no fields used inside it.

So, your change shouldn't cause any problems, but I don't think it's something we have to do.

cehsu:
Thanks for the helpful explanation @jbalboni, closing this.

cehsu:
Thanks for the helpful explanation @jbalboni, the current form allows the user to submit several combinations of values, (commission year and no scholarship info, no commission year and scholarship info). Do we no longer wish to allow this sort of behavior?

jbalboni:
@cehsu I think that behavior is ok and we shouldn't change it.

It looks like I was slightly wrong about why the `seniorROTC` schema works the way it is now. It works because we default the scholarship amounts array to be empty if there are no values. That works differently in RJSF (we won't send an array at all), so removing that required field would probably be helpful for us. Though it might require a backend change.

So, if you want to reopen this with just the required field change, that's fine. Otherwise, we can handle it later when we're submitting the form.

PR #149, opened 2017-09-19
Add new dependent options

rlbaker:
* Adds Father, Mother, Spouse and Other as relationship options in `common/constants.js`
* Rebuilt JSON schemas
* Bumps version to 3.4.0

### DIFF
--- a/dist/10-10EZ-schema.json
+++ b/dist/10-10EZ-schema.json
@@ -226,10 +226,14 @@
         },
         "childRelation": {
           "enum": [
+            "Spouse",
             "Daughter",
             "Son",
             "Stepson",
-            "Stepdaughter"
+            "Stepdaughter",
+            "Father",
+            "Mother",
+            "Other"
           ],
           "type": "string"
         },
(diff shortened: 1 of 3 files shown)


rlbaker:
@jbalboni Do you think I should be renaming the field names to match the fact that it is no longer a "Child" relationship?

jbalboni:
@rlbaker I'm a little hesitant, since it makes this change more complicated. To change the field name we have to:

1. Update the schema to include both field names
2. Update the backend to use dependentRelationship by default, but fall back to childRelationship
3. Update the front end form config to use the new name and write a migration for saved data
4. Update the schema to remove the old name
5. Update the backend to remove the old name

We probably should change it, though, it'll be less confusing later. Unless @lihanli has a strong opinion either way.

lihanli:
i think we should change it, should be low priority though

jbalboni:
Alright, we can make separate tech debt item for that and change it later.

jbalboni:
Debt issue: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4924

PR #153, opened 2017-10-17
Enum to support new HCA eligibility question format

rlbaker:
Adds a new `vaCompensationType` enum field that replaces `isVaServiceConnected`, `compensableVaServiceConnected`, `receivesVaPension` to reflect the new mutually exclusive nature of the fields.

The old fields map to the new enum like so:

### Service-connected disability pay for a 10%, 20%, 30%, or 40% disability rating
Old: `compensableVaServiceConnected: true`
New: `vaCompensationType: 'lowDisability'`

### Service-connected disability pay for a 50% or higher disability rating
Old: `isVaServiceConnected: true`
New: `vaCompensationType: 'highDisability'`

### VA pension
Old: `receivesVaPension: true`
New: `vaCompensationType: 'pension'`

### I don't receive any VA pay
Old: All fields false
New: `vaCompensationType: 'none'`

Currently both styles of fields are supported at the moment. Fallback support will be removed once migrations on the front and backend have been made.

### DIFF
--- a/dist/10-10EZ-schema.json
+++ b/dist/10-10EZ-schema.json
@@ -514,10 +514,10 @@
     "vaCompensationType": {
       "type": "string",
       "enum": [
-        "none",
-        "pension",
         "lowDisability",
-        "highDisability"
+        "highDisability",
+        "pension",
+        "none"
       ]
     },
     "isEssentialAcaCoverage": {
--- a/dist/10-10EZ-schema.json
+++ b/dist/10-10EZ-schema.json
@@ -502,14 +502,14 @@
     "maritalStatus": {
       "$ref": "#/definitions/maritalStatus"
     },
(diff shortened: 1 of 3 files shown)


dist/10-10EZ-schema.json
```diff
@@ -511,6 +511,15 @@
     "receivesVaPension": {
       "type": "boolean"
     },
+    "vaCompensationType": {
+      "type": "string",
+      "enum": [
+        "none",
```
jbalboni: Can we change the order here to:

lowDisability
highDisability
pension
none

It's annoying, but the form library we use doesn't let us re-order enums, so this is the order of the radio buttons you see in the UI.

rlbaker: Done! Thanks.

PR #157, opened 2017-10-30
Simplify pre-needs schema for single applicant flow.

U-DON:


U-DON:
Just getting rid of the `applications` array and flattening its properties down one level.

U-DON:
@mphprogrammer: For `cemeteryNumber` and `highestRank`, is it worth putting in an enum for the list options? At least for `cemeteryNumber`, which still expects a number, but the applicant is not going to know them by number. We should provide names as labels instead.

dist/40-10007-schema.json
```diff
@@ -305,305 +305,299 @@
     }
   },
   "properties": {
-    "applications": {
-      "type": "array",
-      "items": {
```
mphprogrammer: Should we use the schema
```
   "properties": {
      "application": {
        "type": "object",
        "required": { "applicant", "claimant", ....
```

mphprogrammer:
FYI - We have route to get the list of cemeteries:
`GET /v0/preneeds/cemeteries`

dist/40-10007-schema.json
U-DON: The whole thing will be the application, so I don't think there is a point to wrapping everything, unless it's a lot easier to do so for the backend. I can add the required line in either case.

mphprogrammer: I think we should probably do the wrap, if it would be too involved on the FE. Its a resource, and should we resurrect the concept of multiple applications, the structure of { application: ... one or more applications ... } would already be in place - and it does simplify the BE a tad.

U-DON: Done.

mphprogrammer:
LGTM

PR #161, opened 2017-11-03
Update service records for preneeds

U-DON:
Mainly added `nationalGuardState` and also enforced `minItems: 1`. Some minor refactoring.

### DIFF
--- a/dist/40-10007-schema.json
+++ b/dist/40-10007-schema.json
@@ -588,12 +588,12 @@
               "name"
             ],
             "properties": {
+              "name": {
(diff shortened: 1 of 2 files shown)


mphprogrammer:
OK - LGTM, However, we should probably require a nationalGuardState if a national guard (air, army) is selected. There is nothing in the receivePreneedsApplication section of the XSD that requires it, but I've noticed a field returned in EOAS's getBranchesOfService endpoint that says it is (stateRequired).

U-DON:
I can conditionally require it on the form if a National Guard rank is selected.

PR #163, opened 2017-11-10
Update attachments field in preneed schema

jbalboni:
Adds a files array with the attachment id and using the field name from Lihan.

### DIFF
--- a/dist/40-10007-schema.json
+++ b/dist/40-10007-schema.json
@@ -611,18 +611,12 @@
                 "type": "string"
               },
               "attachmentId": {
-                "type": "string",
-                "enum": [
-                  "1",
-                  "2",
-                  "3",
-                  "5",
-                  "6"
-                ]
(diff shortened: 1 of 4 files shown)


lihanli:
there's an api for attachment_types so i assume they can change, maybe we should leave out the enum?

jbalboni:
@lihanli Yeah, I want to ask them about that. Fetching this data from an api is sort of annoying.

dist/40-10007-schema.json
```diff
@@ -596,11 +596,35 @@
...
+                  "1",
+                  "2",
+                  "3",
+                  "5",
+                  "6"
+                ]
+              }
+            },
+            "required": [
+              "attachmentId"
```
lihanli: confirmationCode should be required too

jbalboni:
@lihanli Updated.

PR #164, opened 2017-11-15
Alphabetize serviceBranch labels/format email

cehsu:
Resolves: 
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5862
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5869


src/schemas/40-10007/schema.js
```diff
@@ -13,61 +13,60 @@ _.merge(modifiedToursOfDuty, {
...
-          'RO', // ROTC OF ARMY NAVY OR AF
-          'CA', // US CITIZEN WHO SERVED W/ALLIES
-          'WR', // WOMEN'S RESERVE OF NAVY,MC,CG
-          'CS', // CIVILIAN W/STRATEGIC SVC (OSS)
-          'KC', // QRTRMASTER CORPS KESWICK CREW
-          'CB', // DEFENSE OF BATAAN
-          'CO', // US ARMY TRANSPORT SERVICE
-          'CI', // CIV ID FRIEND/FOE (IFF) TECH
+          'AL', // ALLIED FORCES
           'CC', // AMERICAN FIELD SERVICE
```
cehsu: There was previously duplicated 'American Field Service' labels ('CC' and 'FS'). Should the duplicate be removed?

lihanli: i think it probably shouldn't be removed unless a stakeholder says it should be

src/schemas/40-10007/schema.js
```diff
@@ -119,8 +118,7 @@ definitions.date = {
 
 definitions.email = {
   type: 'string',
-  maxLength: 50,
-  pattern: '[a-zA-Z0-9_.+-]+@[a-zA-Z0-9_+-]+\\.[a-zA-Z]+'
+  format: 'email'
```
jbalboni: I'm not sure we should change this. It may need to be different from the default email format to match a requirement of the backend api.

This also isn't necessary to fix the validation error message on the front end.

src/schemas/40-10007/schema.js
jbalboni: I agree. We should ask about it.

src/schemas/40-10007/schema.js
cehsu: Thanks for pointing that out @jbalboni, this has been reverted.

dist/40-10007-schema.json
```diff
@@ -269,10 +269,6 @@
             "III",
             "IV"
           ]
-        },
-        "maiden": {
```
jbalboni: Should this be somewhere else in the schema?

src/schemas/40-10007/schema.js
cehsu: Sounds good, in that case, it would be good to ask them how the labels should differ. This removal has been reverted.

dist/40-10007-schema.json
cehsu: Good idea @jbalboni. Maiden has been relocated to applicant name.
Also, in looking at how other forms handle this, I noticed that there is an outstanding task for determining the validation for mothersMaidenName in the HCA schema. I could not find a ticket for this, should one be created?

PR #169, opened 2017-11-29
[WIP] Schemas for VR&E Chapter 31 (28-1900) & 36 (28-8832) Forms 

rlbaker:


src/schemas/28-8832/schema.js
```diff
@@ -0,0 +1,119 @@
...
+      $ref: '#/definitions/ssn'
+    },
+    veteranDateOfBirth: {
+      $ref: '#/definitions/date'
+    },
+    veteranServiceBranch: {
+      type: 'string'
+    },
+    veteranDateofDeathMIAPOW: {  // date of death, M.I.A., or P.O.W.
+      $ref: '#/definitions/date'
```
jbalboni: Something to note is that I believe our default date format allows partial dates in the form of `2010-09-XX`. We might need to use the normal date format, depending on api/backend requirements.

src/schemas/28-8832/schema.js
```diff
@@ -0,0 +1,119 @@
...
+    },
+    divorceOrAnnulmentPending: {
+      type: 'boolean',
+    },
+    applicantAgeAtRemarriage: {
+      type: 'number',
```
jbalboni: integer instead, maybe?

src/schemas/28-8832/schema.js
```diff
@@ -0,0 +1,119 @@
...
+    },
+    previousBenefitApplications: {
+      type: 'array',
+      items: {
+        type: 'string'
```
jbalboni: I don't think we have a way to represent a list of strings schema like this on the front end. But we can leave until we know what the UI is supposed to look like.

src/schemas/28-1900/schema.js
```diff
@@ -0,0 +1,114 @@
...
+    },
+    email: {
+      type: 'string',
+      format: 'email'
+    },
+    vaRecordsOffice: {
+      type: 'string',
+    },
+    yearsOfEducation: {
+      type: 'number',
```
jbalboni: Also `integer`, maybe?

src/schemas/28-1900/schema.js
```diff
@@ -0,0 +1,114 @@
...
+    },
+    previousPrograms: {
+      type: 'array',
+      items: {
+        type: 'object',
+        additionalProperties: false,
```
jbalboni: We don't usually use this outside the top level in our schemas (though I think it's set globally whenever we validate, maybe?).

src/schemas/28-8832/schema.js
rlbaker: I wasn't quite sure how this would be getting mapped either, so it's more of a placeholder.

rlbaker:
@jbalboni There's a couple more pieces to add to the 1900 schema that I should have clarified after the meeting tomorrow. I'm also still working on the specs for them as well. I'll get these changes you suggested in today! Thanks for looking it over.

src/schemas/28-1900/schema.js
```diff
@@ -0,0 +1,116 @@
+import constants from '../../common/constants';
+import _ from 'lodash';
+import definitions from '../../common/definitions';
+import schemaHelpers from '../../common/schema-helpers';
+
+let schema = {
+  $schema: 'http://json-schema.org/draft-04/schema#',
+  title: 'DISABLED VETERANS APPLICATION FOR VOCATIONAL REHABILITATION (28-1900)',
+	type: 'object',
```
lihanli: indentation different

PR #170, opened 2017-12-11
Vic schema

lihanli:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6649


dist/definitions.json
```diff
@@ -326,6 +326,17 @@
     "pattern": "^(\\d{4}|XXXX)-(0[1-9]|1[0-2]|XX)-(0[1-9]|[1-2][0-9]|3[0-1]|XX)$",
     "type": "string"
   },
+  "dischargeType": {
+    "type": "string",
+    "enum": [
```
jbalboni: These might end up matching the codes that Preneed uses.

rlbaker: I'm using this in the VR&E schema as well. I'll migrate those to use this once this is merged in.

PR #173, opened 2018-01-15
Update VIC fields

jbalboni:
This updates the schema to match the form design and backend discussions we've had.

### DIFF
--- a/src/schemas/VIC/schema.js
+++ b/src/schemas/VIC/schema.js
@@ -23,8 +23,12 @@ let schema = {
     dd214: Object.assign({}, originalDefinitions.files, {
       minItems: 1,
       maxItems: 1
-    })
-  },
+    }),
+    veteranDateOfBirth: {
+      type: 'string',
+      format: 'date'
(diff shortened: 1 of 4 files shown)


src/schemas/VIC/schema.js
```diff
@@ -17,20 +17,22 @@ let schema = {
...
   },
 };
 
 [
   ['fullName', 'veteranFullName'],
   ['address', 'veteranAddress'],
+  ['ssn', 'veteranSocialSecurityNumber'],
   ['phone'],
-  ['dischargeType']
+  ['date', 'dateOfBirth']
```
lihanli: we don't want XXs in the date for date of birth. also would `veteranDateOfBirth` be a better name since that's what's used in the 1990? do we want to add gender here too or are you using this schema only for the final submission?

jbalboni: I was planning on having gender be one of our `view:` front end only fields. I'll update this for the other two points.

PR #178, opened 2018-02-07
use file definition for photo

lihanli:


dist/VIC-schema.json
```diff
@@ -276,7 +276,23 @@
       "format": "email"
     },
     "photo": {
-      "type": "string"
+      "type": "array",
+      "items": {
+        "type": "object",
```
cehsu: I think we can simply use what's in the items object to define this?

jbalboni: Yeah, either way will work for us, the front end right now is coded as if we're using an object, not an array. But it's not hard to change.

lihanli: ok i changed it

PR #180, opened 2018-02-12
Salesforce states

lihanli:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/8600


jbalboni:
@lihanli Does SF have a list of countries we have to match, too?

lihanli:
@jbalboni yea i added them just now

jbalboni:
@lihanli It looks like they're not actually used in the schema definition, aside from the ones with states.

lihanli:
@jbalboni now it's validating the country

src/schemas/VIC/schema.js
```diff
@@ -1,13 +1,77 @@
...
+      postalCode: {
+        type: 'string',
+        maxLength: 10
+      }
+    }
+  }));
+  countryStateProperties.push({
+    properties: {
+      country: {
+        not: {
```
jbalboni: We could probably get rid of this, right? Since the countriesWithNoState list would skip these?

jbalboni:
@lihanli So can we add back the states that were removed from the US list? I think they're all in the SF settings, just marked as not visible.

lihanli:
@jbalboni ok i added the hidden states from that xml file

PR #183, opened 2018-02-21
Add 526 skeleton

cvalarida:
I don't know what goes in the blanks yet, so this is just a skeleton to make it easier when we do know what shape this should take.

Also, I'm not sure if it even _is_ 21-526b, so...there's that too. Confirmation forthcoming.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "3.22.0",
+  "version": "3.23.0",
   "repository": {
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "3.22.0",
+  "version": "3.23.0",
```
jbalboni: Think you'll need to bump this again.

PR #184, opened 2018-02-21
add employer name

rianfowler:
Employer name is [included](https://app.zenhub.com/workspace/o/department-of-veterans-affairs/vets.gov-team/issues/8823) in the design but not in current 28-1900 [schema](https://github.com/department-of-veterans-affairs/vets-json-schema/blob/master/src/schemas/28-1900/schema.js).

Matches other [form's](https://github.com/department-of-veterans-affairs/vets-json-schema/blob/0e5db3eaa46d9a666ce96fe35ee7d6b71ff77b8b/src/schemas/21P-527EZ/schema.js#L211) approach for capturing employer name.

### DIFF
--- a/dist/28-1900-schema.json
+++ b/dist/28-1900-schema.json
@@ -334,6 +334,9 @@
         "type": "string"
       }
     },
+    "employer": {
+      "type": "string"
+    },
     "monthlyIncome": {
       "type": "number",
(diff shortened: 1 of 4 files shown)


jbalboni:
@rianfowler Needs the version bumped in package.json.

PR #186, opened 2018-02-22
Change education program date fields

jbalboni:
Design uses two year fields instead of a full date.


src/schemas/28-1900/schema.js
```diff
@@ -21,7 +22,7 @@ let schema = {
       type: 'string',
     },
     yearsOfEducation: {
-      type: 'integer',
+      $ref: '#/definitions/year'
```
rianfowler: Is this a count of years?

jbalboni: Bleh, it is.

PR #188, opened 2018-02-23
First pass at 526 fields

cvalarida:
There are still a _lot_ of questions left to answer, but this is what we (think we) know so far.


dist/21-526EZ-schema.json
```diff
@@ -2,7 +2,458 @@
   "$schema": "http://json-schema.org/draft-04/schema#",
   "title": "SUPPLEMENTAL CLAIM FOR COMPENSATION (21-526EZ)",
   "type": "object",
-  "definitions": {},
-  "properties": {},
+  "definitions": {
+    "email": {
```
jbalboni: There's a built in email format for json schema, do you know if you need something different than that?

src/schemas/21-526EZ/schema.js
```diff
@@ -1,9 +1,404 @@
...
+        type: 'string'
+      },
+      attachmentType: {
+        type: 'string'
+      },
+      inflightId: {
+        type: 'string'
+      }
+    },
+    militaryPayments: {
```
jbalboni: Do you need this nesting? It's generally easier for this schema to be as flat as possible and add nesting on the front end.

src/schemas/21-526EZ/schema.js
```diff
@@ -1,9 +1,404 @@
...
+      type: 'object',
+      properties: {
+        payments: {
```
jbalboni: The plural sort of implies there can be more than one, but it's not an array type.

src/schemas/21-526EZ/schema.js
```diff
@@ -1,9 +1,404 @@
...
+      }
+    },
+    directDeposit: {
+      $ref: '#/definitions/directDeposit'
+    },
+    serviceInformation: {
+      type: 'object',
+      properties: {
+        // This is a little different from the common serviceHistory definition
+        servicePeriods: {
```
jbalboni: Should this be an array?

src/schemas/21-526EZ/schema.js
cvalarida: To be honest, I have no idea. This is what it looks like in the excel spreadsheet of the fields. :man_shrugging:

dist/21-526EZ-schema.json
cvalarida: Oh, I didn't know that. I copied this from `10-10EZ`, I think.

src/schemas/21-526EZ/schema.js
cvalarida: Yeah, probably...

src/schemas/21-526EZ/schema.js
cvalarida: No idea. :man_shrugging: 
I'll make it an array and when we run into it, we can change it if needed.

dist/21-526EZ-schema.json
```diff
@@ -2,7 +2,473 @@
...
+            "efctvDt": {
+              "$ref": "#/definitions/date"
+            }
+          }
+        },
+        "homelessness": {
+          "type": "object",
+          "properties": {
+            "hasPointOfContact": {
+              "type": "boolean"
```
omgitsbillryan: This is one of the things I wanted to remove but looks to still be in at the moment.  Seems like the concept of "has point of contact" can be deduced by if `pointOfContact` is `null` or not.

src/schemas/21-526EZ/schema.js
```diff
@@ -1,9 +1,425 @@
...
+    // dateRange: definitions.dateRange // hopefully we can use this later
+    phone: definitions.phone,
+    fullPhone: {
+      type: 'object',
+      properties: {
+        areaNbr: {
+          type: 'string'
+          // Validation?
+        },
+        phoneNbr: {
```
omgitsbillryan: I requested that abbreviations be removed and EVSS agreed.  `areaNbr` ---> `areaNumber` (or `areaCode`???) & `phoneNbr` ---> `phoneNumber`

cvalarida: Excellent! What about `efctvDt`?

cvalarida: Judging by the field `pointOfContactAreaCode`, I think it'd make sense to change `areaNumber` to `areaCode`, but I'll wait until we get confirmation.

cvalarida:
Yeah, this is definitely subject to change. I'd like to get it all ironed out before we do _too_ much work on building the form, but we'll use what we have for now so we're not just waiting around.

PR #189, opened 2018-02-26
Fix invalid schema -- 526

cvalarida:
Missed a spot the first go-round. Woopsadoodle. :grimacing:
I also decided to just call it an array because it only makes sense.


jbalboni:
@cvalarida Need to bump the package.json version.

PR #190, opened 2018-02-28
add dd214 field to 28-1900 schema

rianfowler:
Add discharge document field to schema for this [page](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/8954).

There was no unit test for this field on vic-v2 but I can add one for that validates that at least one object exists.

### DIFF
--- a/src/schemas/28-1900/schema.js
+++ b/src/schemas/28-1900/schema.js
@@ -1,6 +1,7 @@
 import constants from '../../common/constants';
 import _ from 'lodash';
 import definitions from '../../common/definitions';
+import originalDefinitions from '../../common/definitions';
 import schemaHelpers from '../../common/schema-helpers';
 
 let schema = {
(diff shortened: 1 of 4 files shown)


src/schemas/28-1900/schema.js
```diff
@@ -52,6 +53,9 @@ let schema = {
       type: 'number',
       minimum: 0
     },
+    dd214: Object.assign({}, originalDefinitions.files, {
```
jbalboni: Let's change the name of this field to something like `dischargeDocuments`. dd214 is kind of misleading and we may want to fix that on VIC in the future, too.

jbalboni:
@rianfowler Needs package.json version bump.

src/schemas/28-1900/schema.js
rianfowler: Agreed- I'll make a ticket to update it for VIC

PR #192, opened 2018-03-01
Update disability schemas

cehsu:
These changes update the disabilityRating and disabilities schemas


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "3.27.0",
+  "version": "3.30.0",
```
cehsu: Anticipating that the two other VRE schema updates will be merged first.

PR #194, opened 2018-03-08
Reorder treatments

cvalarida:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/8862 had these fields ordered differently.


cvalarida:
Ah, I knew there was a way to make that happen without this, but it's been long enough I completely forgot. I was thinking `updateSchema` or something drastic like that. Thanks!

PR #195, opened 2018-03-16
9091 make military history required

rianfowler:


src/schemas/28-1900/schema.js
```diff
@@ -127,9 +127,21 @@ let schema = {
...
 });
 
+const requiredServiceHistory = _.merge(_.cloneDeep(definitions.serviceHistory), {
+  minItems: 1,
+  items: {
+      required: ['serviceBranch', 'dischargeType'],
+  }
+});
+
+requiredServiceHistory.items.properties.dateRange =  _.merge(_.cloneDeep(definitions.dateRange), {
```
rianfowler: On the fence about this approach- almost just included the entire object in here instead of piecing it from the different parts. Sort of wish i could pass a required argument to `addDefintionToSchema` + `definitions`. I could add another `definitions` called `requiredServiceHistory`.

jbalboni: I think we should add a requiredServiceHistory definition.

rianfowler: Perfect.

rianfowler:
This is ready. Successfully tested this commit on my vets-website branch. @jbalboni

PR #198, opened 2018-04-04
526 use common mg

markgreenburg:
- Replaces as much of the 526 schema as possible with common definitions
- Creates new common definitions where this makes sense (e.g., pciuAddress)


markgreenburg:
I think I've covered all of the points above, but this is my first foray into the vets-json-schema so please tear this PR apart.

src/common/constants.js
```diff
@@ -3981,12 +3981,293 @@ const yesNo = [
 
 const usaStates = _.map(states.USA, (stateData) => { return stateData.value });
```
markgreenburg: FYI, I stole these countries and states by inspecting the `countries` and `states` responses from Letters Confirm Address

src/schemas/21-526EZ/schema.js
```diff
@@ -15,139 +14,118 @@ const directDepositDef = {
...
-      not: {
+const disabilitiesBaseDef = {
+  type: 'array',
+  maxItems: 100,
+  items: {
+    type: 'object',
+    properties: {
+      diagnosticText: {
         type: 'string',
-        'enum': countriesWithStateList
```
markgreenburg: Note that this belongs to the deleted `countryStateProperties` block above the added `disabilitiesBaseDef`; not sure how it got down here in the diff

src/schemas/21-526EZ/schema.js
```diff
@@ -15,139 +14,118 @@ const directDepositDef = {
...
+        type: 'string'
+      },
+      diagnosticCode: {
+        type: 'number'
+      },
+      specialIssueTypeCode: {
+        type: 'string'
+      },
     },
-    state: {
```
markgreenburg: Same as previous comment - this belongs to `countryStateProperties`

src/common/constants.js
```diff
@@ -3981,12 +3981,293 @@ const yesNo = [
...
+  "Vietnam",
+  "Wales",
+  "Western Samoa",
+  "Yemen Arab Republic",
+  "Zambia",
+  "Zimbabwe"
+];
+
+const pciuStates = [
+  "AK",
```
cvalarida: `states.USA` has 62 entries and this has 61. Can we re-use `states.USA` and remove the extra? (Assuming, of course, they map 1:1 otherwise.) If not, we'll have to format this list the same as `states.USA` to give all these options labels for the dropdown menu (and what a pain that'd be!).

dist/definitions.json
```diff
@@ -254,6 +254,365 @@
...
+    ],
+    "required": [
+      "addressLine1",
+      "country"
+    ],
+    "properties": {
+      "addressLine1": {
+        "type": "string",
+        "maxLength": 35,
+        "pattern": "([a-zA-Z0-9-'.,,&#]([a-zA-Z0-9-'.,,&# ])?)+$"
```
cehsu: Should leading carets be added to any of these based on yesterday's meeting?

src/common/definitions.js
```diff
@@ -90,6 +91,86 @@ const address = (() => {
...
+      },
+      zipFirstFive: {
+        type: 'string',
+        pattern: '^\d{5}$' // not in swagger docs
+      },
+      zipLastFound: {
+        type: 'string',
+        pattern: '^\d{4}$' // not in swagger docs
+      },
+      militaryPostOfficeTypeCode: {
```
cvalarida: So you're thinking you want to conditionally hide / require these in the `uiSchema`? I think that makes enough sense.

src/schemas/21-526EZ/schema.js
cvalarida: Git diff is weird. :man_shrugging:

src/schemas/21-526EZ/schema.js
```diff
@@ -15,139 +14,118 @@ const directDepositDef = {
...
-// Not sure this particular kind of datetime definition will be used anywhere else
-// Pattern matches datetimes like 2018-03-22T17:25:19.191Z where the fractional seconds are optional
-// NOTE: This doesn't catch invalid days like February 30th
-const datetime = {
-  pattern: '(\d{4}|XXXX)-(0[1-9]|1[0-2]|XX)-(0[1-9]|[1-2][0-9]|3[0-1]|XX)T([01]\d|2[0-3]):[0-5]\d:[0-5]\d(\.\d+)?Z',
-  type: 'string'
-}
+// Strip suffix from fullName common def
+const fullNameDef = _.merge(definitions.fullName, {
+  properties: _.omit('suffix', definitions.fullName.properties)
```
cvalarida: Oh, right. Because `_.omit()` doesn't accept a path. :man_facepalming:
Good catch.

src/schemas/21-526EZ/schema.js
```diff
@@ -15,139 +14,118 @@ const directDepositDef = {
...
+    address: addressDef,
+    treatmentCenterAddress: treatmentCenterAddressDef,
+    directDeposit: _.merge(definitions.bankAccount, uniqueBankFields),
+    date: definitions.date,
+    dateRange: definitions.dateRange,
+    disabilities: _.merge(disabilitiesBaseDef, {
+      minItems: 1,
+      items: {
+        properties: {
+          secondaryDisabilities: disabilitiesBaseDef
```
cvalarida: Very nice!

src/common/constants.js
markgreenburg: Oh good catch. Yeah, I'll see if I can clean this up

src/common/definitions.js
markgreenburg: I think so. Not having put together any uiSchemas, my opinion is liable to change at any time 🙃

cehsu:
Do any tests need to be updated/added?

src/schemas/21-526EZ/schema.js
```diff
@@ -373,80 +327,7 @@ let schema = {
       required: ['servicePeriods', 'servedInCombatZone']
     },
     disabilities: {
```
cehsu: If this property has the same name as the definition, is it necessary?

src/schemas/21-526EZ/schema.js
```diff
@@ -459,26 +340,11 @@ let schema = {
...
-            pattern: '[a-zA-Z]{2}'
-          },
-          treatmentCenterCity: {
-            type: 'string',
-            maxLength: 100,
-            pattern: "([a-zA-Z0-9\-'.#]([a-zA-Z0-9\-'.# ])?)+$"
+          treatmentCenterAddress: {
+            $ref: '#/definitions/treatmentCenterAddress'
           },
           treatmentCenterType: {
```
cehsu: It looks like there are two options here 'VA_MEDICAL_CENTER', 'DOD_MTF'. Does this only refer to treatments at VA centers? How should we format private center record request data?

src/schemas/21-526EZ/schema.js
```diff
@@ -15,139 +14,118 @@ const directDepositDef = {
...
+      items: {
+        properties: {
+          secondaryDisabilities: disabilitiesBaseDef
         }
       }
-    },
+    }),
+    fullName: fullNameDef,
+    // vets-api will split into separate area code & phone number fields
+    phone: Object.assign({}, definitions.phone, { maxLength: 10 }),
```
cehsu: Should the maxLength be 11? Does the minLength need to be updated as well (default is 10)?

src/schemas/21-526EZ/schema.js
cehsu: Is this being based off alternateName?

src/schemas/21-526EZ/schema.js
```diff
@@ -15,139 +14,118 @@ const directDepositDef = {
...
-    }
-  }
-}));
-countryStateProperties.push({
-  properties: {
-    country: {
-      not: {
+const disabilitiesBaseDef = {
+  type: 'array',
+  maxItems: 100,
```
cehsu: Should we add `minItems: 1` here?

src/schemas/21-526EZ/schema.js
```diff
@@ -15,139 +14,118 @@ const directDepositDef = {
...
+  items: {
+    type: 'object',
+    properties: {
```
cehsu: some of these properties are marked as optional, does that mean the rest are required? 
Including:
- diagnosticText
- disabilityActionType
- decisionCode
- specialIssues
- ratedDisabilityId

cehsu: I see, required is below, would it be possible to move it above?

cehsu: Also, do any additional properties need to be added to the required array?

src/schemas/21-526EZ/schema.js
markgreenburg: It _should_ get a `minItems: 1` property through the merging that's happening in the `definitions` section. I wanted to leave it out of the base definition because this is one place where the `disabilities` and `secondaryDisabilities` schemas differ 😞

cehsu: Thanks for the insight.

src/schemas/21-526EZ/schema.js
cehsu: Actually it looks like the updated length is [7](https://github.com/department-of-veterans-affairs/vets.gov-team/blob/5b9e12c30318acb7d8543fa93e54c210b5bdfd9a/Products/Disability%20Compensation/engineering/EVSS/ResponseTo_questions_for_EVSS_WRT_Form_526_API.md)?

src/schemas/21-526EZ/schema.js
cvalarida: Right, yeah. That's the only name field in the form (which still seems odd to me).

src/schemas/21-526EZ/schema.js
cvalarida: Oh yeah, look at that. :thinking: Does that mean `treatments` is only _meant_ to hold VAMCs? In which case, where do we put private medical records...

Great point, @cehsu.

src/schemas/21-526EZ/schema.js
markgreenburg: Looks like we're missing `disabilityActionType` based on the Swagger:
```
Disability {
diagnosticText (string)stringMax. Length:255Reg. Exp.:([a-zA-Z0-9\-'.,#]([a-zA-Z0-9\-',.# ])?)+$,
disabilityActionType (string) = ['NONE', 'NEW', 'SECONDARY', 'INCREASE', 'REOPEN']stringEnum:"NONE", "NEW", "SECONDARY", "INCREASE", "REOPEN",
decisionCode (string),
specialIssues (Array[SpecialIssue], optional)objectMax. Items:100,
ratedDisabilityId (string),
ratingDecisionId (string, optional),
diagnosticCode (integer, optional),
specialIssueTypeCode (string, optional),
secondaryDisabilities(Array[SecondaryDisability], optional)objectMax. Items:100
}
```

cvalarida: Oh, good point. I'm _guessing_ it'll always be `INCREASE` for this form, but it's worth checking. Also, we should add that in either the front or back end if it's always going to be the same.

@goldenmeanie As the resident SME(ish), do you have any insight on if the `disabilityActionType` will always be `INCREASE` for this form?

src/schemas/21-526EZ/schema.js
```diff
@@ -15,139 +14,118 @@ const directDepositDef = {
...
+ * @property {object} serviceHistory
+ * @param {definitions} definitions the common schema definitions file
+ * @returns {object} the servicePeriods schema object
+ */
+const servicePeriodsDef = ((definitions) => {
+  const serviceHistory = _.cloneDeep(definitions.serviceHistory);
+  serviceHistory.minItems = 1;
+  serviceHistory.maxItems = 100;
+  serviceHistory.items.required = ['serviceBranch', 'dateRange'];
+  delete serviceHistory.items.properties.dischargeType
```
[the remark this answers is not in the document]
markgreenburg: The thinking was that we'd switch our usage of from / to dates to all just use`dateRange`, and require the whole date range. @cvalarida mentioned that we've allowed things with future end dates or 'present' end dates before, which would probably fit our use-case. Thoughts?

src/schemas/21-526EZ/schema.js
markgreenburg: Yeah, that's the only usage. Though checking swagger, it looks as though we don't have the right regex patterns set up to validate against. Will add.

src/schemas/21-526EZ/schema.js
cehsu: Thanks for clarifying.

src/schemas/21-526EZ/schema.js
cehsu: It sounds like in the future this form will support other types?

goldenmeanie: Yes, this is just increase but All Claims will incorporate New and Secondary. I'm not sure about Reopen (but signs point to yes).

src/schemas/21-526EZ/schema.js
markgreenburg: yep, looks like min / max length should be set to 7. Thanks for the link!

src/schemas/21-526EZ/schema.js
markgreenburg: You mean the `$ref`? Yeah, I dunno to be honest. Looks like this is what we do in some of the other forms, e.g. 527EZ (`additionalSources: { $ref: '#/definitions/additionalSources' }`). It's probably more consistent with JSON schema style. 🤷‍♂️

src/schemas/21-526EZ/schema.js
markgreenburg: I believe it refers to both (i.e., both types of centers will use the same fields with the same schema / validations). That's the way it is in the Swagger, at least.

src/schemas/21-526EZ/schema.js
cvalarida: Yeah, I think we still need the property even if we define the definition.

src/schemas/21-526EZ/schema.js
cvalarida: Er, by "VAMCs," I meant VAMCs and DoD MTFs...
Are you saying this might also be used to hold private facilities as well?

src/schemas/21-526EZ/schema.js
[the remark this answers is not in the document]
cvalarida: :thinking: I think you're right, @cehsu. If we make the `dateRange` required, I believe an empty object will satisfy that requirement without satisfying what we're _actually_ looking for. That said, requiring the individual dates is done in the `uiSchema` for the common definition, so I think we're still covered.

PR #199, opened 2018-04-06
526 update semver mg

markgreenburg:
Didn't update version after patch updates to 526 schema. This fixes that problem.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "3.36.1",
+  "version": "3.37.0",
   "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "3.36.0",
+  "version": "3.36.1",
```
cvalarida: Don't know how much it matters at this point, but it might be worth bumping the minor version instead of just the patch. :man_shrugging:

markgreenburg: Yeah, I flip-flopped on this (hah, look at the commit history), but your opinion makes me second-guess myself. I'll change it to a minor release.

PR #201, opened 2018-04-14
[WIP] Update 526

markgreenburg:
526 still doesn't quite match Swagger, need to do more updates in conjunction with building the 526 form.

### DIFF
--- a/src/schemas/21-526EZ/schema.js
+++ b/src/schemas/21-526EZ/schema.js
@@ -105,20 +105,25 @@ const fullNameDef = ((definitions) => {
 })(definitions);
 
 /**
- * Strip address lines from PCIU address def for use with treatment center addresses
+ * Modifies PCIU Address for use with treatments schema
  * @typedef {object} definitions
  * @property {object} pciuAddress
  * @param {definitions} definitions from the common schema definitions file
  * @returns {object} the treatmentCenterAddress schema object
  */
-const treatmentCenterAddressDef = ((definitions) => {
-  const treatmentAddressProperties = _.pick(
-    ['city', 'state', 'country'],
-    definitions.pciuAddress.properties
-  );
-
-  return _.merge(definitions.pciuAddress, { properties: treatmentAddressProperties });
+const treatmentCenterAddressDef = (({ pciuAddress }) => {
+  const { type, oneOf, properties } = pciuAddress;
+  return Object.assign({}, {
+    type,
+    oneOf: _.cloneDeep(oneOf),
+    required: ['country'],
+    properties: {
+      city: Object.assign({}, properties.city)
+    }
+  });
 })(definitions);
+
+console.log(treatmentCenterAddressDef);
 
 let schema = {
   $schema: 'http://json-schema.org/draft-04/schema#',
--- a/dist/21-526EZ-schema.json
+++ b/dist/21-526EZ-schema.json
@@ -669,61 +669,13 @@
         }
       ],
       "required": [
-        "addressLine1",
         "country"
       ],
       "properties": {
-        "addressLine1": {
-          "type": "string",
-          "maxLength": 35,
-          "pattern": "([a-zA-Z0-9-'.,,&#]([a-zA-Z0-9-'.,,&# ])?)+$"
-        },
-        "addressLine2": {
-          "type": "string",
-          "maxLength": 35,
-          "pattern": "([a-zA-Z0-9-'.,,&#]([a-zA-Z0-9-'.,,&# ])?)+$"
-        },
-        "addressLine3": {
-          "type": "string",
-          "maxLength": 35,
-          "pattern": "([a-zA-Z0-9-'.,,&#]([a-zA-Z0-9-'.,,&# ])?)+$"
-        },
         "city": {
           "type": "string",
           "maxLength": 35,
           "pattern": "([a-zA-Z0-9-'.#]([a-zA-Z0-9-'.# ])?)+$"
-        },
-        "zipFirstFive": {
-          "type": "string",
-          "pattern": "^d{5}$"
-        },
-        "zipLastFound": {
-          "type": "string",
-          "pattern": "^d{4}$"
-        },
-        "militaryPostOfficeTypeCode": {
-          "type": "string",
-          "enum": [
-            "APO",
-            "DPO",
-            "FPO"
-          ]
-        },
-        "militaryStateCode": {
-          "type": "string",
-          "enum": [
-            "AA",
-            "AE",
-            "AP"
-          ]
-        },
-        "type": {
-          "type": "string",
-          "enum": [
-            "DOMESTIC",
-            "MILITARY",
-            "INTERNATIONAL"
-          ]
         }
       }
     },
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "3.38.0",
+  "version": "3.39.0",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/department-of-veterans-affairs/vets-json-schema.git"


src/common/definitions.js
```diff
@@ -152,11 +166,16 @@ const pciuAddress = (() => {
         maxLength: 35,
         pattern: "([a-zA-Z0-9\-'.#]([a-zA-Z0-9\-'.# ])?)+$"
       },
+      state: {
+        type: 'string',
+        'enum': usaStates,
+        enumNames: usaStatesLabels
+      },
       zipFirstFive: {
```
cehsu: The UI for this is one field, are we planning to split ZIP up on the FE before submission?

src/common/definitions.js
```diff
@@ -112,16 +118,20 @@ const pciuAddress = (() => {
...
-          .map(state => state.value)
+        enum: usaStates
       }
     }
   }, {
     properties: {
       country: {
+        not: {
+          type: 'string',
+          enum: [USA]
```
cehsu: This is good as is, one idea is whether we should we make the constant an array that includes 'USA', perhaps named USAOnly or somesuch, so that it is consistent with how pciuCountries is used.

src/schemas/21-526EZ/schema.js
```diff
@@ -105,19 +105,40 @@ const fullNameDef = ((definitions) => {
...
  * @returns {object} the treatmentCenterAddress schema object
  */
-const treatmentCenterAddressDef = ((definitions) => {
-  const treatmentAddressProperties = _.pick(
-    ['city', 'state', 'country'],
-    definitions.pciuAddress.properties
-  );
+const vaTreatmentCenterAddressDef = (({ pciuAddress }) => {
+  const { type, oneOf, properties } = pciuAddress;
+  return Object.assign({}, {
```
cehsu: If we decide to use lodash, could we use _.merge here?

src/schemas/21-526EZ/schema.js
```diff
@@ -105,19 +105,40 @@ const fullNameDef = ((definitions) => {
...
 
-  return _.merge(definitions.pciuAddress, { properties: treatmentAddressProperties });
+/**
+ * Grab address lines, city and zip from PCIU address common def, then add
+ * country and state properties for 'oneOf' to work properly
+ */
+const privateTreatmentCenterAddressDef = (({ pciuAddress }) => {
+  const { type, oneOf, required, properties } = pciuAddress;
+  
+  return Object.assign({}, {
```
cehsu: Same question as above.

src/schemas/21-526EZ/schema.js
```diff
@@ -371,6 +393,34 @@ let schema = {
         }
       }
     },
+    privateRecordReleases: {
```
cehsu: These validations are probably all a good idea, but perhaps we should make a comment that none of these are actually required, or externally validated as this info will be sent as an ancillary form? We've discussed this a few times so perhaps future maintainers will benefit from this context.

src/schemas/21-526EZ/schema.js
```diff
@@ -371,6 +393,34 @@ let schema = {
...
+          treatmentDateRange: {
+            $ref: '#/definitions/dateRange'
+          },
+          treatmentCenterAddress: {
+            $ref: '#/definitions/privateTreatmentCenterAddress'
+          },
+          privateMedicalRecordsReleaseAccepted: {
+            type: 'boolean'
+          },
+          'view:privateMedicalRecordsReleasePermissionRestricted': {
```
cehsu: 'view:...' type objects never gets sent to the backend, so we do not need to include it here, we could however, include its contents, namely `privateMedicalRecordsReleasePermissionRestricted`, which should be a boolean.

src/common/definitions.js
markgreenburg: Swagger for the submit endpoint lists these as separate fields (both optional). I hadn't really put enough thought into this, but I figure we'll probably want to do whatever `vets-api` will support. My preference would be to submit as one field since that's what's closest to what we need in the front end (i.e., I think we'd only want users to fill out one zip field that accepts the right pattern for all 5 to 9 digits and a dash). I'll reach out to them and get confirmation of what their plan is.

cehsu: Thanks for clarifying, very helpful @markgreenburg .

src/common/definitions.js
markgreenburg: You're right, I had been using this in its non-array form elsewhere but that's not true anymore.

src/schemas/21-526EZ/schema.js
markgreenburg: I think `merge` merges all properties of a given object from both source and destination which wasn't my intention. That being said, this needs updating anyway since `country` and `state` are things that we need in the VA address and are now included in the common PCIU definition. So 🎉

src/schemas/21-526EZ/schema.js
markgreenburg: Good call

src/schemas/21-526EZ/schema.js
markgreenburg: 👍 this was a holdover from the transferring the form properties straight into the schema. Since this needs to get submitted as part of the separate ancillary form submission I'll make it not a `view` field.

cehsu: Oh does this refer to the warning that gets displayed when `privateMedicalRecordsReleaseAccepted` is selected? We can simply remove it if so.

cehsu: The naming here is slightly confusing, perhaps `privateMedicalRecordsReleaseAccepted` should be `privateMedicalRecordsReleaseRestricted`

markgreenburg: FYI this only controls the text in the conditional alert dialog when a user chooses to restrict their private records request. The boolean that stores whether or not this option is selected is `privateMedicalRecordsReleaseAccepted`

markgreenburg: Ah missed your other comments :)
Agree that the naming is confusing. I'll try to give these properties names closer to their intent.

src/common/definitions.js
```diff
@@ -132,33 +142,42 @@ const pciuAddress = (() => {
...
+        'enum': usaStates,
+        enumNames: usaStatesLabels
       },
       zipFirstFive: {
         type: 'string',
-        pattern: '^\d{5}$' // not in swagger docs
+        pattern: '^\\d{5}$' // not in swagger docs
       },
-      zipLastFound: {
+      zipLastFour: {
```
cvalarida: :man_facepalming: Nice catch.

src/schemas/21-526EZ/schema.js
```diff
@@ -105,19 +105,42 @@ const fullNameDef = ((definitions) => {
...
-const treatmentCenterAddressDef = ((definitions) => {
-  const treatmentAddressProperties = _.pick(
-    ['city', 'state', 'country'],
-    definitions.pciuAddress.properties
-  );
+const vaTreatmentCenterAddressDef = (({ pciuAddress }) => {
+  const { type, oneOf, properties } = pciuAddress;
+  return Object.assign({}, {
+    type,
+    oneOf: oneOf.map((obj) => _.cloneDeep(obj)),
```
cvalarida: Why clone?

src/schemas/21-526EZ/schema.js
```diff
@@ -105,19 +105,42 @@ const fullNameDef = ((definitions) => {
...
+/**
+ * Grab address lines, city and zip from PCIU address common def, then add
+ * country and state properties for 'oneOf' to work properly
+ */
+const privateTreatmentCenterAddressDef = (({ pciuAddress }) => {
+  const { type, oneOf, required, properties } = pciuAddress;
+  
+  return Object.assign({}, {
+    type,
+    oneOf: oneOf.map((obj) => _.cloneDeep(obj)),
```
cvalarida: :clone-wars:

src/schemas/21-526EZ/schema.js
markgreenburg: 🤷‍♂️ Ran into a nasty bug when I accidentally modified the base schema and now I'm paranoid. I'm happy to change this if you think it makes a difference.

PR #208, opened 2018-05-11
526 add back dateRange

markgreenburg:
- Updates zip refs to zipCode to better match patterns used in other forms
- Adds back dateRange to unbreak downstream form config
- Updates minor version number
- Rebuilds dist

### DIFF
--- a/dist/21-526EZ-schema.json
+++ b/dist/21-526EZ-schema.json
@@ -481,7 +481,7 @@
         },
         "zipCode": {
           "type": "string",
-          "pattern": "^\\d{5}(?:[-\\s]\\d{4})?$"
+          "pattern": "^\\d{5}(?:([-\\s]?)\\d{4})?$"
         },
         "militaryPostOfficeTypeCode": {
           "type": "string",
@@ -1444,7 +1444,7 @@
         },
         "zipCode": {
           "type": "string",
-          "pattern": "^\\d{5}(?:[-\\s]\\d{4})?$"
+          "pattern": "^\\d{5}(?:([-\\s]?)\\d{4})?$"
         }
       }
     },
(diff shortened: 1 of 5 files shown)


dist/21-526EZ-schema.json
```diff
@@ -1441,6 +1441,10 @@
             "Philippine Islands",
             "U.S. Minor Outlying Islands"
           ]
+        },
+        "zipCode": {
+          "type": "string",
+          "pattern": "^\\d{5}(?:[-\\s]\\d{4})?$"
```
cvalarida: What do you think about making the separator `[-\s]` optional (`[-\\s]?`)? It's weird, but it would give the user the most freedom and we don't _really_ care at the end of the day, do we? We'll just be stripping out any separator and splitting it up by the first five and last four.

If you don't want to make that change here, that's totally fine by me. It was just a passing thought.

src/schemas/21-526EZ/schema.js
```diff
@@ -158,6 +158,7 @@ let schema = {
     privateTreatmentCenterAddress: privateTreatmentCenterAddressDef,
     directDeposit: _.merge(definitions.bankAccount, uniqueBankFields),
     date: definitions.date,
+    dateRange: definitions.dateRange,
```
cvalarida: Is this because it was still being used somewhere when it was removed?

markgreenburg: Yeah...turns out it's still being used in the `servicePeriodsDef` on L85

dist/21-526EZ-schema.json
```diff
@@ -1441,6 +1441,10 @@
             "Philippine Islands",
             "U.S. Minor Outlying Islands"
           ]
+        },
+        "zipCode": {
```
cehsu: Would it be of interest to create a definition for zipCode and reuse it in the various addresses, perhaps in a separate ticket?

PR #209, opened 2018-05-17
526 service history mg

markgreenburg:
- Updates the `servicePeriods` schema with enums instead of free text.


src/schemas/21-526EZ/schema.js
```diff
@@ -331,7 +309,34 @@ let schema = {
...
+                  'Army',
+                  'Army Reserves',
+                  'Coast Guard',
+                  'Coast Guard Reserves',
+                  'Marine Corp',
+                  'Marine Corps Reserves',
+                  'Navy',
+                  'Navy Reserves',
+                  'NOAA',
+                  'USPHS',
```
cehsu: Should we spell these acronyms out and provide the acronyms in parentheses instead?

markgreenburg: Good call. I'm going to double check with Alex.

src/schemas/21-526EZ/schema.js
```diff
@@ -331,7 +309,34 @@ let schema = {
...
+                  'National Oceanic and Atmospheric Administration',
+                  'Navy',
+                  'Navy Reserve',
+                  'Public Health Service'
+                ]
+              },
+              dateRange: {
+                $ref: '#/definitions/dateRangeAllRequired'
+              }
+            }
```
cvalarida: Should we keep the `minItems`, `maxItems`, and required fields, or were these causing issues in pre-fill?

markgreenburg: Yes, we should definitely keep them. 🤦‍♂️

PR #210, opened 2018-05-18
disallow space-only input on required fields

rianfowler:
My regex skills are about 3/10 and there are a lot of ways to do this so LMK if I should use a different pattern.

Tested that this works. I will add error messages when I do the website PR.
screen shot 2018-05-18 at 09 36 15

Decided to go the schema validation route for 
- client / server app validation parity 
- using schema validation is the most well supported error pattern for accessibility concerns in the client app

I dug a little into json schema documentation to see if there was an easy way to reuse this pattern but didn't find anything.

### DIFF
--- a/dist/10-10EZ-schema.json
+++ b/dist/10-10EZ-schema.json
@@ -171,6 +171,7 @@
             },
             "provinceCode": {
               "type": "string",
+              "pattern": "^.*\\S.*",
               "maxLength": 51
             }
           }
@@ -179,6 +180,7 @@
       "properties": {
         "street": {
           "type": "string",
+          "pattern": "^.*\\S.*",
           "minLength": 1,
           "maxLength": 50
         },
@@ -192,6 +194,7 @@
         },
         "city": {
           "type": "string",
+          "pattern": "^.*\\S.*",
           "minLength": 1,
           "maxLength": 51
         },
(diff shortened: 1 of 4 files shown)


test/schemas/10-10EZ/schema.spec.js
```diff
@@ -79,6 +79,17 @@ describe('healthcare-application json schema', () => {
...
+
+    it('doesn\'t allow names with only spaces', () => {
+      expect(fullNameValidation({ first: '   ', last: '  ' })).to.be.false;
+    });
+  });
+
+  describe('address', () => {
+    const addressValidation = definitionValidator('address');
+    it('doesn\'t allow street, cities, or provinces with only spaces', () => {
+      expect(addressValidation({ street: '   ', city: '    ', country: '     ', provinceCode: '     ' })).to.be.false;
```
rianfowler: These are sort of brittle tests because this validation can pass if any of the validation rules are invalid.

dist/10-10EZ-schema.json
```diff
@@ -171,6 +171,7 @@
             },
             "provinceCode": {
               "type": "string",
+              "pattern": "^.*\\S.*",
```
rianfowler: since this goes into `json` the `\` must be escaped.

src/schemas/10-10EZ/schema.js
```diff
@@ -56,6 +59,8 @@ let schema = {
         },
         city: {
           type: 'string',
+          pattern: '^.*\\S.*',
+          minLength: 2,
```
jbalboni: Seems like we have duplicate minLength properties in a few places

PR #212, opened 2018-05-30
Preneeds validation fixes

lihanli:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10648


jbalboni:
@lihanli Their state field has a max length of 3, and their postal code has a max of 5 (but I think the latter is going to have to change).

PR #214, opened 2018-06-05
526 homelessness mg

markgreenburg:
- Adds updated homelessness schema to this repo (currently it's defined in the form config).
- Updates relevant validation patterns to latest found in evss swagger.

Requires another PR to switch to using this schema in `vets-website`.

### DIFF
--- a/dist/21-526EZ-schema.json
+++ b/dist/21-526EZ-schema.json
@@ -2225,24 +2225,20 @@
         "homelessness": {
           "type": "object",
           "required": [
-            "hasPointOfContact"
+            "isHomeless"
           ],
           "properties": {
-            "hasPointOfContact": {
-              "type": "boolean",
-              "default": false
+            "isHomeless": {
+              "type": "boolean"
             },
             "pointOfContact": {
               "type": "object",
-              "required": [
-                "pointOfContactName",
-                "primaryPhone"
-              ],
               "properties": {
                 "pointOfContactName": {
                   "type": "string",
+                  "minLength": 1,
                   "maxLength": 100,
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "3.49.0",
```
cehsu: When do we decide to go to 3.49.1 vs 3.50.0?

jbalboni: It has consequences as far as what upgrade commands you need to run to update versions, but it doesn't make much difference for us since we're referencing the git commit hash anyway. We should follow semvar, but mainly the version just needs to change whenever we need to pull a new version into vets-website.

dist/21-526EZ-schema.json
```diff
@@ -2225,24 +2225,20 @@
...
-                "pointOfContactName",
-                "primaryPhone"
-              ],
               "properties": {
                 "pointOfContactName": {
                   "type": "string",
+                  "minLength": 1,
                   "maxLength": 100,
-                  "pattern": "^([a-zA-Z0-9\\-'.#][a-zA-Z0-9\\-'.# ]?)*$"
+                  "pattern": "^([a-zA-Z0-9-/]+( ?))*$"
```
cvalarida: Why update this? Is this the updated pattern in the swagger docs?

markgreenburg: Yeah, exactly. I guess they changed it on the sly. 🤷‍♂️

cvalarida: Bugger. Looks like we're gonna have to do _another_ thorough sweep of the whole bloody schema to make sure it matches the swagger docs before we're done.

PR #216, opened 2018-06-07
hca: Missing Required Field: Insurance policy number or group code.

lihanli:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10858


jbalboni:
@lihanli Needs a package.json update, I think.

dist/10-10EZ-schema.json
```diff
@@ -272,18 +272,18 @@
       "properties": {
         "first": {
           "type": "string",
-          "pattern": "^.*\\S.*",
           "minLength": 1,
-          "maxLength": 30
+          "maxLength": 30,
+          "pattern": "^.*\\S.*"
```
rianfowler: Does this order matter?

src/schemas/10-10EZ/schema.js
```diff
@@ -42,12 +42,11 @@ let schema = {
       type: 'object',
       oneOf: countryStateProperites,
       properties: {
-        street: {
+        street: Object.assign({
```
rianfowler: This would probably look better using the spread operator.

dist/10-10EZ-schema.json
lihanli: order of object keys?

rianfowler: Yeah was just wondering if this was changed on purpose or just some undone edits.

lihanli: the order shouldn't matter, order of object keys isn't guaranteed in js

PR #221, opened 2018-06-18
526 use new address schema mg

markgreenburg:
* Removes PCIU common def
* Adds new, much simpler address schema for 526 addresses, fwding addresses directly within 526 schema
* Removes private record release schema (done in this PR because it referenced the old PCIU common address def, and we don't need the schema at all anymore since we're not building 4142s for veterans)
* Updates some constants for state values
* Updates build files & version number


src/schemas/21-526EZ/schema.js
```diff
@@ -36,9 +40,46 @@ const disabilitiesBaseDef = {
...
-
-// Extracted to enable easy adding of properties for forwardingAddress
-const addressDef = definitions.pciuAddress;
+const addressBaseDef = {
+  type: 'object',
+  required: ['country', 'city', 'addressLine1'],
+  properties: {
+    country: {
+      type: 'string',
+      'enum': pciuCountries,
```
cehsu: Are we interested in using 'United States' as the label for USA?

src/schemas/21-526EZ/schema.js
```diff
@@ -36,9 +40,46 @@ const disabilitiesBaseDef = {
...
+      default: 'USA'
+    },
+    addressLine1: {
+      type: 'string',
+      maxLength: 35,
+      pattern: "([a-zA-Z0-9-'.,,&#]([a-zA-Z0-9-'.,,&# ])?)+$"
```
cvalarida: Should we perhaps prefix these patterns with `^` even though EVSS doesn't? Also...because this is living in vets-website, I think we should clean up those extra `,`s as well. (As much as I'd like to do more, I think we should probably leave the refactoring after that.)

src/schemas/21-526EZ/schema.js
markgreenburg: Yeah, this has gone back and forth a few times. I confirmed with Tim that EVSS only accepts 'USA', and it seems kind of silly to create a value / label map for just one country. Not ideal, but we do it elsewhere across the site, and I've set the default to 'USA' so at least US-based veterans don't have to go hunting through the list.

src/schemas/21-526EZ/schema.js
markgreenburg: Yeah, that's a good call.

PR #222, opened 2018-07-02
add schema for 22-0993

jholton:


jholton:
Just noticed that it looks like I need to add "VA file number":
https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Products/Colmery%20Act%20Opt%20Out/Design/Colmery_optout_outline.md

src/schemas/22-0993/schema.js
```diff
@@ -0,0 +1,37 @@
...
+  type: 'object',
+  definitions: _.pick(definitions, [
+    'fullName',
+    'ssn',
+    'vaFileNumber',
+    'privacyAgreementAccepted'
+  ]),
+  additionalProperties: false,
+  properties: {},
+  oneOf: [
```
jholton: Is 'oneOf' good, or should I use 'anyOf'?

jholton:
added "VA file number"

src/schemas/22-0993/schema.js
lihanli: are people allowed to enter both?

jholton: good question.  The doc I linked to says:

* Social security number [number input] [required]
* I don't have an SSN [checkbox] [optional]
  * VA file number [revealed and required if I don't have an SSN is checked]

So, theoretically they would only enter one, but I can't think of a scenario where it would hurt if they entered both for some reason.  I'll change it to 'anyOf'

src/schemas/22-0993/schema.js
```diff
@@ -0,0 +1,37 @@
...
+
+let schema = {
+  $schema: 'http://json-schema.org/draft-04/schema#',
+  title: 'REQUEST TO OPT-OUT OF INFORMATION SHARING WITH EDUCATIONAL INSTITUTIONS (VA FORM 22-0993)',
+  type: 'object',
+  definitions: _.pick(definitions, [
+    'fullName',
+    'ssn',
+    'vaFileNumber',
+    'privacyAgreementAccepted'
```
lihanli: not needed, `schemaHelpers.addDefinitionToSchema` will add each definition, empty object should work for this field

lihanli:
can you add as spec and update the version in package.json?

PR #223, opened 2018-07-11
22-0993: change 'veteran' to 'claimant'

jholton:


dist/22-0993-schema.json
```diff
@@ -52,10 +52,10 @@
   },
   "additionalProperties": false,
   "properties": {
-    "veteranFullName": {
+    "claimantFullName": {
```
cehsu: IIRC we often use applicant if not veteran, is claimant the new generic standard?

cehsu: Actually it looks like we use both "claimant" and "applicant", are there any guidelines on when to use one or the other?

jholton: Good question.  I just used 'claimant' because I had seen that before.  I hadn't personally interacted with any of the schemas that use 'applicant'.  I don't have a preference.

jholton: I'm going to go ahead and merge this.  If we come up with a standard on 'applicant' vs 'claimant' it can easily be changed.

PR #224, opened 2018-07-11
change regex to get rid of warning in vets-api

jholton:
When running the vet-api test suite, a number of warnings are displayed:
`gems/json-schema-2.7.0/lib/json-schema/attributes/pattern.rb:10: warning: character class has '-' without escape: /^([a-zA-Z0-9-'.,&#]([a-zA-Z0-9-'.,&# ])?)+$/`

altering these regexes fixes it.

### DIFF
--- a/dist/21-526EZ-schema.json
+++ b/dist/21-526EZ-schema.json
@@ -232,22 +232,22 @@
         "addressLine1": {
           "type": "string",
           "maxLength": 35,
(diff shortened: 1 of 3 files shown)


jholton:
[StackOverflow reference](https://stackoverflow.com/questions/24843335/rails-regex-warning-character-class-has-without-escape/24843354#24843354)

PR #228, opened 2018-07-18
Complaint tool schema

cehsu:
These changes set up the complaint tool schema.

Outstanding questions have been noted as PR comments, and as inline comments, including those relating to discrepancies between the provided list, the example request, and the design (e.g. level of study).

TODO:
- [x] tests
- [x] bump version
- [x] build dist

### DIFF
--- a/src/schemas/complaint-tool/schema.js
+++ b/src/schemas/complaint-tool/schema.js
@@ -1,0 +1,478 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "GI BILL SCHOOL COMPLAINT TOOL",
+  "type": "object",
+  "additionalProperties": false,
+  "definitions": {
+    "address": {
+      "type": "object",
+      "oneOf": [
+        {
+          "properties": {
+            "country": {
+              "type": "string",
+              "enum": [
+                "CAN"
+              ]
+            },
+            "state": {
+              "type": "string",
+              "enum": [
+                "AB",
+                "BC",
+                "MB",
+                "NB",
+                "NF",
+                "NT",
+                "NV",
+                "NU",
+                "ON",
+                "PE",
+                "QC",
+                "SK",
+                "YT"
+              ]
+            },
+            "postalCode": {
+              "type": "string",
+              "maxLength": 10
+            }
+          }
+        },
+        {
+          "properties": {
+            "country": {
+              "type": "string",
+              "enum": [
+                "MEX"
+              ]
+            },
+            "state": {
+              "type": "string",
+              "enum": [
+                "aguascalientes",
+                "baja-california-norte",
+                "baja-california-sur",
+                "campeche",
+                "chiapas",
+                "chihuahua",
+                "coahuila",
+                "colima",
+                "distrito-federal",
+                "durango",
+                "guanajuato",
+                "guerrero",
+                "hidalgo",
+                "jalisco",
+                "mexico",
+                "michoacan",
+                "morelos",
+                "nayarit",
+                "nuevo-leon",
+                "oaxaca",
+                "puebla",
+                "queretaro",
+                "quintana-roo",
+                "san-luis-potosi",
+                "sinaloa",
+                "sonora",
+                "tabasco",
+                "tamaulipas",
+                "tlaxcala",
+                "veracruz",
+                "yucatan",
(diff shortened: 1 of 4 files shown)


src/schemas/complaint-tool/schema.js
```diff
@@ -0,0 +1,171 @@
...
+      activeICN: {
+        type: 'number'
+      },
+      historicalICN: { // TODO: should this be "historicalICNs" (plural)?
+        type: 'array',
+        items: {
+          type: 'object',
+          properties: {
+            ICN: {
+              type: 'number'
```
cehsu: Are these typically type number?

src/schemas/complaint-tool/schema.js
```diff
@@ -0,0 +1,171 @@
...
+      type: 'string',
+      maxLength: 1000
+    }
+  }
+};
+
+[
+  ['privacyAgreementAccepted'], // should this be removed on the BE?
+  ['usaPhone', 'phone'],
+  ['address'],
```
cehsu: Do we want to copy how VIC defines address here?

src/schemas/complaint-tool/schema.js
```diff
@@ -0,0 +1,171 @@
...
+      ]
+    },
+    fullName: {
+      '$ref': '#/definitions/fullName'
+    },
+    email: {
+      type: 'string',  // TODO: determine if there is a length limit
+      format: 'email'
+    },
+    profileData: {
```
lihanli: we don't need to include this in the schema, it won't be used in the frontend

src/schemas/complaint-tool/schema.js
```diff
@@ -0,0 +1,171 @@
...
+            'Mrs.',
+            'Ms.',
+            'Dr.',
+            'Other'
+          ]
+        }
+      }
+    })
+  },
+  required: [
```
cehsu: @lihanli how do we typically handle conditional requirements? For example name is only required if "anonymous" is not selected? Is this only handled on the FE? [Schema draft 7 includes if else support](https://ajv.js.org/keywords.html#ifthenelse), should we consider using that?

lihanli: something like this should work
```javascript
{
  oneOf: [
    {
      properties: {
        onBehalfOf: {
          type: 'string',
          'enum': [
            'Myself',
            'Someone else'
          ]
        },
      }
      required: ['fullName']
    },
    {
      properties: {
        onBehalfOf: {
          type: 'string',
          'enum': [
            'Anonymous'
          ]
        },
      }
    }
  ]
}
```

cehsu: Thanks @lihanli .

src/schemas/complaint-tool/schema.js
```diff
@@ -0,0 +1,165 @@
...
+      format: 'email'
+    },
+    educationDetails: {
+      required: [ // no fields are required for submission, though several are required by design on FE
+        // 'school',
+        // 'programs'
+      ],
+      school: definitions.school,
+      programs: { // TODO: Needs to be translated into a (comma) delimited string, and spelled out if within 255 limit 
+        type: 'object',
```
rianfowler: Maybe could use an array of enum here.

cehsu: Thanks @rian, these will be presented as checkboxes, which we typically create using booleans.

src/schemas/complaint-tool/schema.js
```diff
@@ -0,0 +1,165 @@
...
+      type: 'string',
+      maxLength: 1000
+    }
+  }
+};
+
+[
+  ['privacyAgreementAccepted'],
+  ['usaPhone', 'phone'], // TODO: note validation requirements pending feedback
+  ['address'], // TODO: note length limits pending feedback
```
lihanli: did we get an answer about international addresses? seems like we should use the same address definition that's used in the vic schema

cehsu: Actually, only US addresses will be accepted, both the school and the applicant address have different validations as well.

cehsu:
Outstanding questions, a list of translations has been added to the [wiki](https://github.com/department-of-veterans-affairs/vets.gov-team/wiki/Rainbows-2018-Technical-Documentation).

PR #229, opened 2018-07-25
Issue/11714

cehsu:
These changes incorporate the latest round of stakeholder feedback as documented [here](#11714). These changes also set up the [schema](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11499).

### DIFF
--- a/dist/complaint-tool-schema.json
+++ b/dist/complaint-tool-schema.json
@@ -39,97 +39,141 @@
   ],
   "properties": {
     "address": {
-      "type": "object",
-      "required": [
-        "street",
-        "city",
-        "state",
-        "postalCode",
-        "country"
+      "type": "string",
+      "minLength": 1,
+      "maxLength": 499
+    },
+    "address2": {
+      "type": "string",
+      "minLength": 1,
+      "maxLength": 499
+    },
+    "city": {
+      "type": "string",
+      "minLength": 1,
+      "maxLength": 25
+    },
+    "state": {
+      "type": "string",
+      "enum": [
+        "AL",
+        "AK",
+        "AZ",
+        "AR",
+        "CA",
+        "CO",
+        "CT",
+        "DE",
+        "DC",
+        "FL",
+        "GA",
+        "HI",
+        "ID",
+        "IL",
+        "IN",
+        "IA",
+        "KS",
+        "KY",
+        "LA",
+        "ME",
+        "MD",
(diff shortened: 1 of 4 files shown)


src/schemas/complaint-tool/schema.js
```diff
@@ -83,38 +82,35 @@ let schema = {
...
-        postalCode: {  // TYPE: text (5)
-          type: 'string',
-          pattern: '^\\d{5}$' // common definition pattern (meets submission requirements)
-        },
-        country: { // TODO: determine validation, picklist? 
-          type: 'string',
-          'enum': ['US'] // Only US addresses are supported
-        }
-      }
+    // TRANSLATE: address and address2 must be combined into a single comma separated string property named 'address'
```
erikphansen: would it be useful to add an example of what the final combined `address` should look like given sample values for the input `address` and `address2` fields? Such as "`given 'address': '123 Main St.' and 'address2': 'APT 23' then the final translated 'address' should be '123 Main St., APT 23'`

dist/complaint-tool-schema.json
```diff
@@ -39,97 +39,141 @@
...
+      ]
+    },
+    "postalCode": {
+      "type": "string",
+      "pattern": "^\\d{5}$"
+    },
+    "country": {
+      "type": "string",
+      "enum": [
+        "US"
```
rianfowler: Some [schools](url
https://www.vets.gov/gi-bill-comparison-tool/profile/31100792) are not in the US

cehsu: Thanks @rianfowler, stakeholders indicated that supporting international addresses should be revisited in the future. We can add a comment about that here, and double check with them if of interest.
Another thought relating to this is whether we should display the country field at all if there is only one choice allowed, and simply set this on the BE. But perhaps that is a question for design, and I will ask on the contact page PR.

src/schemas/complaint-tool/schema.js
cehsu: @lihanli are we planning to perform these translations on the BE? If so, would anything in addition to these "TRANSLATE" comments be helpful?

lihanli: an example should be fine

cehsu: Great idea @erikphansen, updated.

PR #235, opened 2018-07-31
526 update uploads mg

markgreenburg:
Updates the uploads schema for 526 so that uploads work again.

### DIFF
--- a/src/common/constants.js
+++ b/src/common/constants.js
@@ -4207,6 +4207,34 @@ const pciuStates = states.USA
   .concat(statesOnlyInPCIU)
   .sort((stateA, stateB) => (stateA.label - stateB.label));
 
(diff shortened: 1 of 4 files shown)


src/common/constants.js
```diff
@@ -4207,6 +4207,34 @@ const pciuStates = states.USA
   .concat(statesOnlyInPCIU)
   .sort((stateA, stateB) => (stateA.label - stateB.label));
 
+  const documentTypes526 = [
+    { value: 'L015', label: 'Buddy/Lay Statement' },
```
cvalarida: Where does this list come from again?

markgreenburg: 😬 @SMLuthi dug it up for me.

PR #237, opened 2018-08-07
Update feedback tool to support international school addresses

cehsu:


dist/complaint-tool-schema.json
```diff
@@ -427,12 +427,247 @@
                     "country": {
                       "type": "string",
                       "enum": [
-                        "US"
+                        "AFG",
```
cehsu: @lihanli @rianfowler do we have updated validation information for school address country?

src/schemas/complaint-tool/schema.js
```diff
@@ -189,39 +193,38 @@ let schema = {
...
-              address: {
+              address: { // TRANSLATE: all international address fields into street, street2, and country
                 type: 'object',
                 required: ['street', 'city', 'state', 'country', 'postalCode'],
                 properties: {
                   street: {
                     type: 'string',
                     minLength: 1,
-                    maxLength: 126 // address + address2 length must be < 255
+                    maxLength: 126 // street + street2 length must be < 255 (INTL 80 on FE)
```
cehsu: These maxLengths add up to 249, with 5 left over for spaces and commas to be inserted. Does this breakdown of limits between seem reasonable?

We previously discussed using `street`, `street2`, `city`, and `country` for international addresses (and only removing `state` and `postal code` from international addresses), but putting `state` and `postal code` onto `street2`, but leaving `city` in the `city` field, seems like it would be unclear to read, and these limits are set up with the idea that we will also move `city` into `street2` in mind.

src/schemas/complaint-tool/schema.js
```diff
@@ -68,6 +72,86 @@ const fullName = _.set('properties.suffix', {
...
+      type: 'string',
+      'enum': [USA].map(country => country.value),
+      enumNames: [USA].map(country => country.label),
+      default: USA.value
+    }
+  }
+};
+
+// TRANSLATE: all international address fields into street, street2, and country
+const internationalSchoolAddress = {
```
cehsu: These maxLengths add up to 249, with 5 left over for spaces and commas to be inserted. Does this breakdown of limits between seem reasonable?

We previously discussed using `street`, `street2`, `city`, and `country` for international addresses (and only removing `state` and `postal code` from international addresses), but putting `state` and `postal code` onto `street2`, but leaving `city` in the `city` field, seems like it would be unclear to read, and these limits are set up with the idea that we will also move `city` into `street2` in mind.

rianfowler: Seems reasonable

PR #245, opened 2018-08-20
Delete old properties

markgreenburg:
Removes a couple of old properties so that our form submits without `vets-api` validation errors (which were being thrown because these two old properties happened to be required in the schema).

Tested locally by having vets-api use the new schema and the submission went through.

### DIFF
--- a/dist/21-526EZ-schema.json
+++ b/dist/21-526EZ-schema.json
@@ -1516,8 +1516,7 @@
     "serviceInformation": {
       "type": "object",
       "required": [
-        "servicePeriods",
-        "servedInCombatZone"
+        "servicePeriods"
       ],
       "properties": {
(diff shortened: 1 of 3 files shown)


dist/21-526EZ-schema.json
```diff
@@ -1634,8 +1629,7 @@
       "items": {
         "type": "object",
         "required": [
-          "treatmentCenterName",
-          "treatmentCenterType"
```
jbalboni: Looks like there's some code referencing this in the front end, but it looks like it's removing this property, so it shouldn't break or anything.

markgreenburg: That's a good point, I can take out that `omit` when I open a PR to update the version number.

PR #246, opened 2018-08-22
4142 auth to disclose info

Charles-Stone:
Included in this PR is the schema for 4142, test for the 4142 schema and an updated Package.json with the version number pushed to the next number.


robfelder:
@department-of-veterans-affairs/code-reviewers-for-external-teams-group - Good Morning, we are looking to get a code review on the this new Schema.

lihanli:
can you change the indentation to 2 spaces? that's what we're using for all the other files

src/schemas/21-4142/schema.js
```diff
@@ -0,0 +1,42 @@
...
+            type: 'string'
+        },
+    },
+    required: ['privacyAgreementAccepted', 'claimantFullName']
+};
+
+[
+    ['fullName', 'claimantFullName'],
+    ['ssn', 'veteranSocialSecurityNumber'],
+    ['centralMailVaFile', 'veteranVaFileNumber'],
```
lihanli: we have been using `vaFileNumber` for the name of this field on other forms and generally try to keep the naming consistent unless there's a good reason not to

test/schemas/21-4142/schema.spec.js
```diff
@@ -0,0 +1,44 @@
...
+
+    // Veteran Phone
+    sharedTests.runTest('phone', ['applicantPrimaryPhone']);
+
+    // Veteran Email
+    sharedTests.runTest('email', ['claimantEmail']);
+});
+
+//Limited Consent
+schemaTestHelper.testValidAndInvalid('limitedConsent', {
```
lihanli: shouldn't this be nested in the '21-4142 schema' describe block?

cvalarida:
@robfelder @Charles-Stone Hey! :wave:

I'll second what @lihanli and @erikphansen said, but other than that, it looks good to me.

To build, you can run `npm run build`. That'll run all the schemas through and output them as JSON to `dist/`, which is where we import them from when we go to use them in vets-website (and I think vets-api, but I'm fuzzy on that one).

For the version bump, we typically increment the minor version. I think this is for...something something Ruby gem. I'm a front end engineer, so again, a little fuzzy there. :man_shrugging:

erikphansen:
If someone wants to weigh in on the reason the version bump happens (other than as a best practice since adding a schema *is* a version change), it would be educational for me!

src/schemas/21-4142/schema.js
```diff
@@ -0,0 +1,78 @@
...
+    },
+    veteranDateOfBirth: {
+      $ref: '#/definitions/date'
+    },
+    veteranAddress: {
+      $ref: '#/definitions/address'
+    },
+    email: {
+      type: 'string',
+    format: 'email'
```
erikphansen: nit: indentation

test/schemas/21-4142/schema.spec.js
```diff
@@ -0,0 +1,61 @@
+import SchemaTestHelper from '../../support/schema-test-helper';
+import schemas from '../../../dist/schemas';
+import fixtures from '../../support/fixtures';
+import _ from 'lodash';
+import SharedTests from '../../support/shared-tests';
+
+const schema = schemas['21-4142'];
+
+const schemaDefaults = {
+    privacyAgreementAccepted: true
```
lihanli: indentation

src/schemas/21-4142/schema.js
```diff
@@ -0,0 +1,78 @@
...
+  definitions: _.pick(definitions, [
+    'fullName',
+    'ssn',
+    'vaFileNumber',
+    'date',
+    'address',
+    'phone',
+    'dateRange',
+    'privacyAgreementAccepted'
+  ]),
```
lihanli: any reason why you stopped using `schemaHelpers.addDefinitionToSchema`? it will add the definition here and add the field in `properties`

lihanli:
@erikphansen the version bump happens iirc because the frontend npm library won't auto update unless the version has changed

erikphansen:
@lihanli I don't think that's right. We bring in the json schema directly from github based on a commit hash https://github.com/department-of-veterans-affairs/vets-website/blob/issue/12558/package.json#L237

Maybe we don't *need* to bump the version on this. Not that I think we should stop doing that!

lihanli:
@erikphansen @jbalboni first mentioned it here https://github.com/department-of-veterans-affairs/vets-json-schema/pull/49#issuecomment-272232030, maybe he has more information

jbalboni:
@erikphansen It used to be the case with `npm` that if you didn't bump the version it wouldn't pull down a new version of the library when someone ran `npm install` later, even if the github commit had been changed. So we would see errors in Jenkins because someone changed the github commit, but not the version, and the build server would use the old version.

 I don't know if this is the case with `yarn`, though.

erikphansen:
Weird stuff, @jbalboni. Thanks for the details.

src/schemas/21-4142/schema.js
Charles-Stone: I'll resolve.

test/schemas/21-4142/schema.spec.js
Charles-Stone: I'll resolve

src/schemas/21-4142/schema.js
Charles-Stone: I can change this back to using addDefinitionToSchema for the handful of fields. This was the 8th time I had written the schema in an attempt to follow best practices and ensure the tests passed. In doing so this was just where I ended up with the implementation.

Charles-Stone:
Changes requested have been made and push to the current version of 4142-auth-to-disclose-info branch.

src/schemas/21-4142/schema.js
```diff
@@ -0,0 +1,67 @@
...
+import _ from "lodash";
+
+
+let schema = {
+  $schema: 'http://json-schema.org/draft-04/schema#',
+  title: 'AUTHORIZATION TO DISCLOSE INFORMATION TO THE DEPARTMENT OF VETERANS AFFAIRS (21-4142/21-4142a)',
+  type: 'object',
+  additionalProperties: false,
+  definitions: _.pick(definitions, [
+    'address',
```
lihanli: this line is not needed

src/schemas/21-4142/schema.js
```diff
@@ -0,0 +1,67 @@
...
+    'phone',
```
lihanli: can you use `schemaHelpers.addDefinitionToSchema` for phone?

test/schemas/21-4142/schema.spec.js
```diff
@@ -0,0 +1,61 @@
+import SchemaTestHelper from '../../support/schema-test-helper';
+import schemas from '../../../dist/schemas';
+import fixtures from '../../support/fixtures';
+import _ from 'lodash';
+import SharedTests from '../../support/shared-tests';
+
+const schema = schemas['21-4142'];
+
+const schemaDefaults = {
```
lihanli: `schemaDefaults` are not needed when you are deleting the required fields below

src/schemas/21-4142/schema.js
```diff
@@ -0,0 +1,67 @@
...
+  required: ['privacyAgreementAccepted', 'veteranFullName']
+};
+
+[
+  ['privacyAgreementAccepted'],
+  ['fullName', 'veteranFullName'],
+  ['ssn', 'veteranSocialSecurityNumber'],
+  ['vaFileNumber'],
+  ['date', 'veteranDateOfBirth'],
+  ['address', 'veteranAddress']
```
lihanli: this form is going to be converted into PDF and submitted to the central mail API right? you should use the special `centralMailAddress` and `centralMailVaFile` definitions. when making a pdf submission request the API also requires the file number and zip code to be submitted as part of the `metadata` and they have to be formatted in a certain way. the normal address definition can be used for `providerFacilityAddress` since they are probably expecting the zipcode from `veteranAddress`.

lihanli:
can you update the version number?

Charles-Stone:
@lihanli I've made the requested changes.

lihanli:
@Charles-Stone looks like the version number still needs to be changed, and `schemaDefaults` are not needed for the spec

Charles-Stone:
@lihanli requested changes made.

PR #254, opened 2018-09-06
Add another valid schema for school addresses

erikphansen:
- Also fix nesting of education details.

Update tests

[This PR](https://github.com/department-of-veterans-affairs/vets-website/pull/8327) adds code that sends data to the backend using this new address schema


src/schemas/FEEDBACK-TOOL/schema.js
```diff
@@ -92,7 +92,50 @@ const internationalSchoolAddress = {
   }
 };
 
-const schoolAddress = [domesticSchoolAddress, internationalSchoolAddress];
+// TRANSLATE: all search tool address fields into street, street2, and country
+const searchToolSchoolAddress = {
+  required: ['viaSearchTool', 'street', 'city', 'country'],
+  properties: {
+    viaSearchTool: {
```
cehsu: This is actually not necessary to specify, looping in @jholton to clarify if possible. As long as the data satisfies one of the schemas, it will be considered valid, it is not necessary to specify which schema should be used for the validation.
Also, if we plan to include information in the schema that is not intended to be sent back (besides using `view:` fields on the frontend), sometimes it's helpful to add a comment in the schema, to call out that we're proposing the backend perform a translation, to remove that information.

erikphansen: I think we actually do need it. Imagine a case where the country is 'United States' but we don't pass in all the data that's required for a US address, such as `postalCode`. Then it will fail the validation for a domestic address but still pass the new validation I just added. So without adding this additional flag we won't be able to enforce the additional requirements for US addresses.

src/schemas/FEEDBACK-TOOL/schema.js
```diff
@@ -92,7 +92,50 @@ const internationalSchoolAddress = {
...
+      type: 'string', // TYPE: text (255)
+      minLength: 1,
+      maxLength: 255
+    },
+    street: {
+      type: 'string',
+      minLength: 1,
+      maxLength: 80
+    },
+    street2: {
```
cehsu: It seems like we should add a third street field to the manual and search international school address schemas, since the school search results feature addresses with three address lines, so as to provide a consistent experience for the user (manual entry supporting the same number of fields as the search tool). We can adjust the line length limits (80 --> 53), and include this as another translation that the backend can perform. Looping in @jholton for the backend perspective on this translation idea.

src/schemas/FEEDBACK-TOOL/schema.js
```diff
@@ -92,7 +92,50 @@ const internationalSchoolAddress = {
...
+    },
+    postalCode: {
+      type: 'string',
+      minLength: 1,
+      maxLength: 30
+    }
+  }
+}
+
+const schoolAddress = [domesticSchoolAddress, internationalSchoolAddress, searchToolSchoolAddress];
```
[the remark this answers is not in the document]
erikphansen: Maybe? I just left it how it was, but if making this plural is the best practice I'll change it.

src/schemas/FEEDBACK-TOOL/schema.js
erikphansen: So add a third street field to manual entry only for international addresses?

src/schemas/FEEDBACK-TOOL/schema.js
cehsu: We will need to specify which domestic address schema we are using on the frontend, so there is no risk of the wrong domestic schema being used to validate a particular field.

src/schemas/FEEDBACK-TOOL/schema.js
cehsu: Good question, we could/should also do this for domestic, that would be more consistent with the user experience we typically provide.

src/schemas/FEEDBACK-TOOL/schema.js
erikphansen: How would we do that unless the schemas have different shapes? That's what I was trying to accomplish with adding `viaSearchTool`.

cehsu: Specific types of schemas are selected from oneOf lists on the frontend as:
```
const domesticSchoolAddress = schoolAddress.oneOf[0];
const internationalSchoolAddress = schoolAddress.oneOf[1];
```
https://github.com/department-of-veterans-affairs/vets-website/blob/b93fdd14b3dcaea712a3b3aade48499190d666a5/src/applications/edu-benefits/feedback-tool/config/form.js#L40-L41

src/schemas/FEEDBACK-TOOL/schema.js
erikphansen: Okay. That's easier. I just don't know when I've been presented with three different fields for street address before :) I guess that's more common outside of the US.

src/schemas/FEEDBACK-TOOL/schema.js
erikphansen: Okay, I need to process this a bit.

erikphansen: @cehsu Other than prevent invalid input on form fields, is there another layer of frontend validation? I'm asking because picking a school via the search tool bypasses all of the form input validation.

cehsu: incorrectly prefilled/fetched data prevents users from advancing to the next page (e.g. the service branch, and contact info address country bugs) on the frontend, and prevents submission to the backend, i believe, @jholton does that sound correct?

jholton: Yes, if the submitted form does not pass schema validation on the backend, the api will raise an exception and return a 422.

src/schemas/FEEDBACK-TOOL/schema.js
jholton: @cehsu We can transform as necessary to fit the Salesforce requirements.

src/schemas/FEEDBACK-TOOL/schema.js
erikphansen: @cehsu After looking at this a lot more, I'm still thinking the method I came up with is a valid choice, but I have some questions about what the existing code is still doing. First, I noticed that the two school address schemas that are brought in are used, but all of their [required attributes are stripped out](https://github.com/department-of-veterans-affairs/vets-website/blob/search-tool-school-address/src/applications/edu-benefits/feedback-tool/config/form.js#L44) so all the validation is done via `ui:required`s. I'm not sure why that is. ~~I'm also not sure why the schema for school address is [updated here](https://github.com/department-of-veterans-affairs/vets-website/blob/search-tool-school-address/src/applications/edu-benefits/feedback-tool/config/form.js#L409) if all of the form fields stay the same whether it's a domestic or international address.~~ (I think I figured that one out. It has to do with some of the string lengths being different for domestic vs intl)

When using the school search tool, I'm not sure we even can do validation on the front end since the tool is simply setting the school address data directly. I still need to look into that a bit more.

src/schemas/FEEDBACK-TOOL/schema.js
```diff
@@ -67,12 +72,17 @@ const internationalSchoolAddress = {
...
-      maxLength: 80
+      maxLength: 53
     },
     street2: {
       type: 'string',
       minLength: 1,
-      maxLength: 80
+      maxLength: 53
+    },
+    street3: {
```
cehsu: Depending on how we combine these on the backend, we may want to reduce these to 52, to allow for an additional ", " (comma and space), between street2 and street3.

src/schemas/FEEDBACK-TOOL/schema.js
cehsu: Good point, a user shouldn't encounter a validation error using the school search tool. Is it possible to determine that the data coming in will validate properly? Should we set up error logging in case we encounter fields that are too long? Should we also create an abbreviation helper for strings that are too long (e.g. replace street with ST).

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "3.88.0",
+  "version": "3.90.0",
```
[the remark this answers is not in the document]
erikphansen: d'oh! yeah.

src/schemas/FEEDBACK-TOOL/schema.js
erikphansen: I had a similar thought after submitting this. I think that abbreviation helpers would be a bit heavy-handed. I'm assuming that the backend will completely reject the submission if a field is too long and thus fails the JSON schema validation check. If that's the case, I think we should just truncate fields that are too long. Even if the entire address is not captured in Salesforce, having the fist ~53 characters in each street field should be more than enough info for someone to know which institution is being reported.

PR #256, opened 2018-09-12
Enforce uppercase schema names in dist

cehsu:
These changes enforce uppercase schema names in the dist folder.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "3.89.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "3.89.0",
+  "version": "3.90.0",
```
erikphansen: This one probably doesn’t need a version bump, but it won’t hurt anything either.

PR #257, opened 2018-09-14
[WIP] BAH-Sprint 3 - Add form 21-0781 to definitions and the 526EZ schema

Charles-Stone:
This is a WIP PR to support internal review of Form 781 & 781a schema creation (#12701)

As a developer on Claims Modernization I need for the form 781 schema to be created.

Assumptions: The 781 and 781a schema will be integrated with the 526 schema in the same way the 4142 schema is incorporated with the 526 schema.

### DIFF
--- a/dist/21-526EZ-schema.json
+++ b/dist/21-526EZ-schema.json
@@ -1667,6 +1667,9 @@
     },
     "form4142": {
       "$ref": "#/definitions/form4142"
+    },
+    "form0781": {
+      "$ref": "#/definitions/form0781"
     }
   }
 }
--- a/dist/definitions.json
+++ b/dist/definitions.json
@@ -1007,5 +1007,80 @@
         "$ref": "#/definitions/privacyAgreementAccepted"
       }
     }
+  },
+  "form0781": {
+    "type": "object",
+    "incident": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "personalAssault": {
+            "type": "boolean"
+          },
+          "medalsCitations": {
+            "type": "string"
+          },
+          "incidentDate": {
+            "$ref": "#/definitions/date"
+          },
+          "incidentLocation": {
+            "type": "string"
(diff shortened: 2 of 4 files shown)


abhijitvk:
@Charles-Stone : Based on the call with AdHoc team this afternoon, there are couple of items that were discovered which will impact this PR:

- Form 21-0781/0781a is completed only with a new, original claim, or a new, non-original claim. The current  src/schemas/21-526EZ/schema.js is used only for "Claims for Increase" form. Plan is to create a new schema for 526 All Claims later in the near future.
- In the current development process, a local version of schema.js file is used in 526 All Claims. Location of the file is : https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/applications/disability-benefits/all-claims/config/schema.js
- AdHoc suggested that we modify the above "schema.js" file for any 781 / 781a / 4142 fields

**To summarize for this PR:**

      - Revert 0781 form field changes done in the file "src/schemas/21-526EZ/schema.js"
      - Retain the definitions.json changes (LGTM) and push it to the master branch. In the future, we can use the definitions of 4142 and 0781 within new schema of 526 all claims 

@kreek @cvalarida : Please do correct if I have misstated anything

@randyhoffmastergit @robfelder - fyi

cvalarida:
@abhijitvk You're exactly right. I think it makes sense for this to be put directly into the schema file you mentioned. The difference between this and the 4142 is that I don't think we need to put it in the definition file since we won't be using it in both the increase-only app and the all-claims app (or anywhere else).

The schema looks great, though! I'm not sure what any of the validations should be, but I'd imagine there are probably string length limitations and the like.

abhijitvk:
@Charles-Stone : I think this must be closed without any merge? Do you agree, if yes can u close this one out.

abhijitvk:
Closing the PR as no longer needed. 781 schema changes were completed in 526 all claims via vets-website and in the new schema file in this repo

PR #258, opened 2018-09-19
Add SSN last four to GIBFT schema

erikphansen:


dist/FEEDBACK-TOOL-schema.json
```diff
@@ -8,6 +8,10 @@
       "pattern": "^(\\d{4}|XXXX)-(0[1-9]|1[0-2]|XX)-(0[1-9]|[1-2][0-9]|3[0-1]|XX)$",
       "type": "string"
     },
+    "ssnLastFour": {
+      "type": "string",
+      "pattern": "^[0-9]{4}$"
```
cehsu: Are we meant to allow XX (according to the full validation)? Also, would we want consider something like `(?!0000)[0-9]{4}`?

erikphansen: I’m working on more robust validation on the FE. The full SSN validation on the backend is similarly simple.

PR #260, opened 2018-09-20
change 41-10007 schema to reflect xsd

jholton:
Changes based on [latest XSD](https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Products/Pre-Needs-Burial/api/application.v4.xsd)

### DIFF
--- a/src/schemas/40-10007/schema.js
+++ b/src/schemas/40-10007/schema.js
@@ -91,7 +91,8 @@ _.merge(modifiedToursOfDuty, {
         ]
       },
       highestRank: {
-        type: 'string'
+        type: 'string',
+        maxLength: 20
       },
       nationalGuardState: {
         type: 'string',
@@ -124,10 +125,14 @@ const emailFormat = {
   format: 'email'
 };
 
-definitions.address.required = ['street'];
+definitions.address.required = ['street', 'city', 'state', 'postalCode'];
 definitions.address.properties.street.maxLength = 20;
 definitions.address.properties.street2.maxLength = 20;
 definitions.address.properties.city.maxLength = 20;
+definitions.address.oneOf.forEach((obj) => {
+  obj.properties.postalCode.maxLength = 5;
+  obj.properties.state.maxLength = 3;
+});
 
 definitions.date = {
   type: 'string',
@@ -137,6 +142,7 @@ definitions.date = {
 definitions.fullName.properties.first.maxLength = 15;
 definitions.fullName.properties.last.maxLength = 25;
 definitions.fullName.properties.middle.maxLength = 15;
+definitions.fullName.properties.suffix.maxLength = 3;
 
 definitions.phone.minLength = 0;
 definitions.phone.maxLength = 20;
@@ -310,6 +316,9 @@ let schema = {
             properties: {
               attachmentId: {
                 type: 'string'
+              },
+              name: {
+                maxLength: 50
               }
             }
           }
--- a/src/schemas/40-10007/schema.js
+++ b/src/schemas/40-10007/schema.js
@@ -311,7 +311,7 @@ let schema = {
         },
         preneedAttachments: _.merge({}, originalDefinitions.files, {
           items: {
-            required: ['attachmentId', 'confirmationCode'],
+            required: ['attachmentId', 'confirmationCode', 'name'],
             properties: {
               attachmentId: {
                 type: 'string'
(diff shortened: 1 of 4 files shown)


src/schemas/40-10007/schema.js
```diff
@@ -124,10 +125,14 @@ const emailFormat = {
   format: 'email'
 };
 
-definitions.address.required = ['street'];
+definitions.address.required = ['street', 'city', 'state', 'postalCode'];
```
jholton: Will create a new frontend ticket for this one too.  Currently on the frontend, if the user selects a country other than  'USA' or 'Canada', the `state` field disappears.  Submission will fail.  `state` is required no matter what country is selected.

jholton: Actually, I think this is a backend issue.  If no state is present, the backend should populate it with a default value.  Thoughts?  Thoughts on what that default value should be?
I'm going to email EOAS stakeholders too.

src/schemas/40-10007/schema.js
```diff
@@ -124,10 +124,14 @@ const emailFormat = {
...
 definitions.address.properties.street.maxLength = 20;
 definitions.address.properties.street2.maxLength = 20;
 definitions.address.properties.city.maxLength = 20;
+definitions.address.oneOf.forEach((obj) => {
+  obj.properties.postalCode.maxLength = 5;
+  obj.properties.state.maxLength = 3;
```
erikphansen: The Mexican states are longer than three letters. Does this schema need to adjust those?

src/schemas/40-10007/schema.js
jholton: Good question.  I didn't think about it because we currently leave the postalCode on the form as a required field regardless of country selected.

src/schemas/40-10007/schema.js
jholton: Maybe.  We currently don't show a `state` field on the frontend when Mexico is selected.
If we go with a solution of submitting a placeholder/default `state` value for all non-USA/Canada countries, then it probably does not need to change.

If we actually add the state field to the form for Mexico then it will need to change.

I did create a ticket for this issue here:  https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13461

I have emailed EOAS stakeholders for their feedback.

src/schemas/40-10007/schema.js
```diff
@@ -310,6 +315,9 @@ let schema = {
             properties: {
               attachmentId: {
                 type: 'string'
+              },
```
cehsu: can we please add name to the required list?

src/schemas/40-10007/schema.js
```diff
@@ -310,6 +315,9 @@ let schema = {
             properties: {
               attachmentId: {
```
cehsu: Any interest in enumerating the enum and enumNames here?
```
                          'enum': [
                            '1',
                            '2',
                            '3',
                            // '4',
                            '5',
                            '6'
                          ],
                          enumNames: [
                            'Discharge',
                            'Marriage related',
                            'Dependent related',
                            // 'VA preneed form',
                            'Letter',
                            'Other'
```

src/schemas/40-10007/schema.js
erikphansen: Good point. Thanks.

src/schemas/40-10007/schema.js
jholton: Done.

src/schemas/40-10007/schema.js
jholton: Done

src/schemas/40-10007/schema.js
erikphansen: Ah, got it. I was confused about why the Mexican states are even listed in the schema in that case. But that's just the result of us using the common definition. 👍

src/schemas/40-10007/schema.js
erikphansen: could you clarify what you mean here @cehsu?

jholton: The `attachmentId` was just defined as a `string`.
On the form it's a dropdown, so I added the `enum` definition as @cehsu defined it in her comment.

erikphansen: haha, I was looking at your code that had already implemented her feedback so I was confused :)

PR #263, opened 2018-09-24
13553# 4142 schema updates to fix date range, address definition

abhijitvk:
Updates for 4142 form properties

1) Form 4142 property "treatmentDateRange" to be of type array
2) Form 526EZ property "treatments" to have minimum item as 0, to fix the schema validation error for empty "treatments"=>[]
3) For the 4142 property "providerFacilityAddress", use "centralMailAddress" definition to use the correct address properties and avoid conflict with 526 schema definition of "address"

### DIFF
--- a/src/common/definitions.js
+++ b/src/common/definitions.js
@@ -534,13 +534,8 @@ const form4142 = {
           },
           treatmentDateRange: {
             type: 'array',
-            properties: {
-              from: {
-                $ref: '#/definitions/date'
-              },
-              to: {
-                $ref: '#/definitions/date'
-              }
+            items: {
+              $ref: '#/definitions/dateRange'
             }
           },
           providerFacilityAddress: {
--- a/src/common/definitions.js
+++ b/src/common/definitions.js
@@ -533,10 +533,18 @@ const form4142 = {
             type: 'string'
           },
           treatmentDateRange: {
-            $ref: '#/definitions/dateRange'
+            type: 'array',
+            properties: {
+              from: {
+                $ref: '#/definitions/date'
+              },
+              to: {
+                $ref: '#/definitions/date'
+              }
+            }
           },
           providerFacilityAddress: {
-            $ref: '#/definitions/address'
+            $ref: '#/definitions/centralMailAddress'
           }
         }
       }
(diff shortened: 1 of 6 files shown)


abhijitvk:
@Charles-Stone : Could you review the changes from your side? If it looks good, I can get the next set of reviewer assigned to it.

abhijitvk:
@kreek @cvalarida : Can one of you please review the changes and approve it if it looks good for the merge?

abhijitvk:
@kreek : It seems that one of the fix in this PR for "treatments" property is also tracked via https://github.com/department-of-veterans-affairs/vets-json-schema/pull/264. 

If rest of the changes looks fine, can this PR be merged?

src/common/definitions.js
```diff
@@ -533,10 +533,18 @@ const form4142 = {
             type: 'string'
           },
           treatmentDateRange: {
-            $ref: '#/definitions/dateRange'
+            type: 'array',
+            properties: {
```
cvalarida: Is the idea here to make it an array of date ranges? If so, what do you think of calling this property `treatmentDateRanges` (plural to imply an array) and make it something like:
```js
treatmentDateRanges: {
  type: 'array',
  items: {
    $ref: '#/definitions/dateRange'
  }
}
```

Arrays require an `items` property and objects require a `properties` property. I don't think it'll work if we give an array `properties`.

src/schemas/21-526EZ/schema.js
```diff
@@ -143,7 +143,8 @@ let schema = {
     }),
     fullName: fullNameDef,
     phone: definitions.usaPhone,
-    form4142: definitions.form4142
+    form4142: definitions.form4142,
+    centralMailAddress: definitions.centralMailAddress
```
cvalarida: :thinking: I see why this is required, but for posterity, can you add a comment saying this is for the 4142 schema since it's not immediately obvious?

abhijitvk: Added comment for 4142 schema

src/common/definitions.js
abhijitvk: @cvalarida : Thanks for the feedback on the date range. Idea is to have date ranges to match the physical form. However we don't have UI to support date ranges in this phase. We already have the property configured in UI for transformation and in the backend for PDF generation. Based on your feedback, I made the change to use #/definitions/dateRange. Is this acceptable?

PR #265, opened 2018-09-26
686 - `address` changes

jholton:
These are changes to the dependent's place of birth based on documents and comments from Syed.
Here is the logic from the document  "VDC UC005_ Prepare Form 686c"

```
Business Rules – Fill Form Field 14B
1. For form fields 14B-1, 14B-2, 14B-3:
    a.  If 'Child Place of Birth Country' = "United States", then 'Child Place of Birth' = 'Child Place of Birth City' + ", " + 'Child Place of Birth State" (UC05: BR66)
    b.  Else if 'Child Place of Birth Country' = "Not in List", then 'Child Place of Birth' = 'Child Place of Birth Country Name' (UC05: BR67)
    c.  Else then 'Child Place of Birth' = 'Child Place of Birth Country' (UC05: BR68)
```

I organized the schema using the way the [Feedback Tool handles school addresses](https://github.com/department-of-veterans-affairs/vets-json-schema/blob/master/src/schemas/FEEDBACK-TOOL/schema.js#L255) as my example.

To paraphrase how this should work:
```
For a dependents place of birth, the user selects a country.
If user selects 'USA'
  then `city` (text field) and `state` (dropdown) are displayed and are required
If user selects 'Not in list'
  then a text field appears for the user to type the country name.  `city` and `state` are not shown.
If user selects any other country from the list
  then that is all.  `city` and `state` are not shown.
```

The list of states and countries used is subject to change.
I got the list of states from the document `EVSS DAM Form 686c_674 Updated version 17_14.xlsx` file.  This file does not have a list of valid countries.
Syed said there are endpoints that will return the list of states and countries, but I am still waiting on clarification as to what those endpoints are.


src/schemas/21-686C/schema.js
```diff
@@ -37,8 +55,57 @@ let schema = {
           },
           childAddress: schemaHelpers.getDefinition('address'),
           personWhoLivesWithChild: schemaHelpers.getDefinition('fullName'),
+
```
cehsu: extra space.

src/schemas/21-686C/schema.js
```diff
@@ -1,7 +1,25 @@
 import definitions from '../../common/definitions';
+import { countries, states50AndDC } from '../../common/constants';
 import schemaHelpers from '../../common/schema-helpers';
 import _ from 'lodash';
 
+const textRegex = '^(?!\\s)(?!.*?\\s{2,})[^<>%$#@!^&*][\\w ]+$';
```
lihanli: why is this needed?

jholton: Good question.  
From `EVSS DAM Form 686c_674 Updated version 17_14.xlsx`:
```
"For each character:
1. Any alphabetic lower or upper-case character
2.. No more than one consecutive space
3. The following Special Characters are not permitted (<, >, %, $, #, @, !, ^, &, *)
4. No numbers
5. Only valid values should be entered"
```

I have not tried submitting values that violate these rules to see if the service validates them.

jholton:
Added more address related changes.

src/schemas/21-686C/schema.js
```diff
@@ -1,18 +1,482 @@
...
   title: 'DECLARATION OF STATUS OF DEPENDENTS',
   type: 'object',
   additionalProperties: false,
-  definitions: {},
+  definitions,
   properties: {
+    claimantAddress: {
+      type: 'object',
+      oneOf: [domesticAddress, militaryAddress, internationalAddressDropDown, internationalAddressText]
+    },
```
lihanli:
```
{
  type: 'object',
  oneOf: [domesticAddress, militaryAddress, internationalAddressDropDown, internationalAddressText]
}
```
can this be put in a variable since it's used in multiple places?

test/schemas/21-686C/schema.spec.js
```diff
@@ -20,31 +20,151 @@ describe('21-686C schema', () => {
   sharedTests.runTest('fullName', ['veteranFullName', 'claimantFullName']);
   sharedTests.runTest('ssn', ['veteranSocialSecurityNumber', 'claimantSocialSecurityNumber', 'spouseSocialSecurityNumber']);
   sharedTests.runTest('vaFileNumber', ['vaFileNumber', 'spouseVaFileNumber']);
-  sharedTests.runTest('address', ['claimantAddress', 'spouseAddress']);
+  // sharedTests.runTest('address', ['claimantAddress', 'spouseAddress']);
```
lihanli: delete this line?

cehsu:
@jholton we typically infer military address if the state code selected is a military state code (rather than specify military as an option in the country list). Could that work here?

src/schemas/21-686C/schema.js
```diff
@@ -1,18 +1,482 @@
...
+  "Venezuela",
+  "Vietnam",
+  "Wales",
+  "Western Samoa",
+  "Yemen Arab Republic",
+  "Zambia",
+  "Zimbabwe"
+]
+
+const states = [
```
erikphansen: do we not already have this array of states defined elsewhere and available to import?

src/schemas/21-686C/schema.js
```diff
@@ -1,18 +1,482 @@
...
+      maxLength: 50,
+      'enum': [countryUSA, countryMilitary, countryNotInList].concat(countriesForeign)
+    }
+  }
+}
+
+const postalCode = {
+  type: 'string',
+  maxLength: 10,
+  pattern: '^[0-9]{5,10}$'
```
erikphansen: I might be wrong but if the postalCode needs to be 5 or 10 characters, then a `-` is also a valid character since 10 digit zips are in the format: 12345-6789. Maybe this pattern should really be two patterns: one for 5 digit zips and one for zip+4.

jholton: I had similar thoughts about a dash, but forgot to follow up.
In their documentation it is defined as 
```
The maximum amount of characters for this field is 10. Please only enter numeric values.
```

jholton:
> @jholton we typically infer military address if the state code selected is a military state code (rather than specify military as an option in the country list). This will also generally result in the post office type field being displayed. Could that work here?

Ah, I see.  I see now that in `constants` our `states.USA` contains all the states I listed + the military postalTypes.  I can just use that (and add "United States Minor Outlying Islands" to the list).

@erikphansen  ^^ pertinent to your question about the list of states.

jholton:
Pushed changes that include military postalTypes in the list of states.

jholton:
Merging.  Will open new PR for further changes.

PR #269, opened 2018-10-04
686: change references to 'claimant' to 'veteran'

jholton:
Since `claimant` and `veteran` are the same now, removing references to `claimant`

### DIFF
--- a/dist/21-686C-schema.json
+++ b/dist/21-686C-schema.json
@@ -106,7 +106,7 @@
     }
   },
   "properties": {
-    "claimantAddress": {
+    "veteranAddress": {
       "type": "object",
       "oneOf": [
         {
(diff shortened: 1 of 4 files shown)


dist/21-686C-schema.json
```diff
@@ -2962,9 +2959,6 @@
     "veteranSocialSecurityNumber": {
       "$ref": "#/definitions/ssn"
     },
-    "claimantSocialSecurityNumber": {
```
erikphansen: Are these two deletions from this file something that should have been done in a previous PR?

jholton: Not necessarily.  We decided that the claimant will always be the veteran, so no need to have both `claimantSocialSecurityNumber ` and `veteranSocialSecurityNumber ` on the form.

PR #276, opened 2018-10-12
remove unused 686 fields

lihanli:
these fields are not in the evss api


src/schemas/21-686C/schema.js
```diff
@@ -379,13 +379,6 @@ let schema = {
     liveWithSpouse: {
       type: 'boolean'
     },
-    monthlySpousePayment: {
```
erikphansen: good, that answers my question about this field.

PR #278, opened 2018-10-24
Add Form 8940 to definitions and 21-526EZ schema.

Charles-Stone:
Form 8940: Schema creation#14081

As a developer on Claims Modernization, I need for the form 8940 schema to be created.

Supporting Artifact

    https://www.vba.va.gov/pubs/forms/vba-21-8940-are.pdf

Assumptions

    The 8940 schema will be integrated with the 526 All claims schema in the same way the 4142 schema is incorporated with the 526 Claims for increase schema.
    This story is related to #12701 (for reference only)
    The schema is only needed for 8940; not for 4192 since 4192 will only be downloaded/uploaded

Acceptance Criteria

    Schema is available in repo and the unit tests have run successfully
    Schema contains the appropriate data elements
    Unit tests are developed and runs successfully

### DIFF
--- a/dist/definitions.json
+++ b/dist/definitions.json
@@ -1014,5 +1014,169 @@
         "$ref": "#/definitions/privacyAgreementAccepted"
       }
     }
+  },
+  "form8940": {
+    "type": "object",
+    "unemployability": {
+      "type": "object",
+      "properties": {
+        "disabilityPreventingEmployment": {
+          "type": "string"
+        },
+        "underDoctorHopitalCarePast12M": {
+          "type": "boolean"
+        },
+        "doctorProvidedCare": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "name": {
+              "type": "string"
+            },
+            "address": {
+              "$ref": "#/definitions/address"
+            },
+            "dates": {
+              "$ref": "#/definitions/dateRange"
+            }
+          }
+        },
+        "hospitalProvidedCare": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "name": {
+              "type": "string"
+            },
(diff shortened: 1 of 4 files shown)


dist/definitions.json
```diff
@@ -1014,5 +1014,169 @@
...
+            },
+            "dates": {
+              "$ref": "#/definitions/dateRange"
+            }
+          }
+        },
+        "disabilityAffectedEmploymentFullTimeDate": {
+          "$ref": "#/definitions/date"
+        },
+        "lastWoredFullTimeDate": {
```
abhijitvk: @Charles-Stone typo error. Field name to be "lastWorkedFullTimeDate"

dist/definitions.json
```diff
@@ -1014,5 +1014,169 @@
...
+        },
+        "doctorProvidedCare": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "name": {
+              "type": "string"
+            },
+            "address": {
+              "$ref": "#/definitions/address"
```
abhijitvk: I am unsure for the address what the reqs are for 8940 form. If the schema uses #/definitions/address then as per 526 all claims schema, required fields are [‘country’, ‘city’, ‘addressLine1’]. However if we use #/definitions/centralMailAddress, then postalCode is required unless you have overridden centralMailAddress to use baseAddressDef

dist/definitions.json
```diff
@@ -1014,5 +1014,169 @@
...
+            },
+            "workType": {
+              "type": "string"
+            },
+            "hoursPerWeek": {
+              "type": "string"
```
abhijitvk: type to be "number"

PR #279, opened 2018-10-29
686 schema changes to make compatible with prefill

jholton:


src/schemas/21-686C/schema.js
```diff
@@ -98,6 +98,21 @@ let schema = {
...
+        pattern: '^[-0-9]{10,12}$'
+      },
+      ssn: {
+        oneOf: [
+          {
+            type: 'string',
+            pattern: '^[0-9]{9}$'
+          }, {
+            type: 'string',
+            pattern: '^[0-9]{3}-[0-9]{2}-[0-9]{4}$'
```
lihanli: from what they said in the channel the api allows both types of ssn, how about we allow just the 9 digit ssn so it's consistent across our other forms?

jholton:
Sounds good to me. I made the change because the endpoint returns them with
dashes, but I can easily strip the dashes out when prefilling.

src/schemas/21-686C/schema.js
```diff
@@ -98,6 +98,10 @@ let schema = {
   additionalProperties: false,
   definitions: _.merge(definitions,
     {
+      phone: {
+        type: 'string',
+        pattern: '^[-0-9]{10,12}$'
```
erikphansen: do we not have a more robust regex for phone numbers? seems like this would allow a lot of invalid phone numbers through.

jholton: Yeah, this is a pretty lazy regex.  I'm going to use our predefined `usaPhone` definition (i.e. allows 10 digits) and strip out dashes when prefilling.

PR #306, opened 2018-12-13
Updating 781 schema for all-claims 

dmccurdybah:
This PR contains changes to form781 in the all-claims schema.

New fields:

`additionalIncidentText` - Free text entered by a user who has more than 3 PTSD related incidents (the current maximum number of incidents supported by the all-claims flow).   The data is handled as overflow-data on the 781 form.

`additionalSecondaryIncidentText ` - Free text entered by a user who has more than 3 PTSD-assault related incidents (the current maximum number of incidents supported by the all-claims flow).  The data is handled as overflow-data on the 781a form.

`otherInformation` - An array of strings that populates a text box on the 781a form. Too many lines overflow to additional PDF pages.

Updated fields:

`remarks` - This is being moved from the incident to the `form781` because it is collected once per 781 form.

Other changes:

`form781` properties are now nested correctly under `properties`

`secondaryPtsdIncident.authorities` were renamed to `source` to match the `form781` schema (I think we were expecting there to be multiple types of sources but authorities are the only source.


src/schemas/21-526EZ-allclaims/schema.js
```diff
@@ -594,78 +594,93 @@ const schema = {
...
-                  type: 'string'
-                },
-                injuryDeathDate: {
-                  $ref: '#/definitions/date'
-                },
-                unitAssigned: {
-                  type: 'string'
+      properties: {
+        "remarks": {
+          "type": "string"
```
abhijitvk: remove double quotes from the key and use single quote for the value

src/schemas/21-526EZ-allclaims/schema.js
```diff
@@ -594,78 +594,93 @@ const schema = {
...
+        },
+        "otherInformation": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        incident: {
+          type: 'array',
+          minItems: 1,
```
abhijitvk: Do we need minItems? You can have `form0781: null` as a valid payload

src/schemas/21-526EZ-allclaims/schema.js
dmccurdybah: Thanks!

src/schemas/21-526EZ-allclaims/schema.js
dmccurdybah: I know we talked about this yesterday, and I think you were right.  The minItems will only be enforced when we have form781 element, which will not happen unless we have an incident.

dist/21-526EZ-ALLCLAIMS-schema.json
```diff
@@ -2498,78 +2498,93 @@
...
-                  "type": "string"
-                },
-                "injuryDeathDate": {
-                  "$ref": "#/definitions/date"
-                },
-                "unitAssigned": {
-                  "type": "string"
+      "properties": {
+        "remarks": {
+          "type": "string"
```
abhijitvk: need to re-generate this file via npm run build

dmccurdybah: There weren't any changes that needed to be pushed up due to a rebuild.

dmccurdybah:
@cvalarida and @markgreenburg - These changes are related to the all-claims transform of 781 data, so I'll have a related `vets-website` PR as well.

src/schemas/21-526EZ-allclaims/schema.js
```diff
@@ -594,78 +594,93 @@ const schema = {
...
+              incidentDescription: {
+                type: 'string'
+              },
+              unitAssigned: {
+                type: 'string'
+              },
+              unitAssignedDates: {
+                $ref: '#/definitions/dateRange'
+              },
+              personInvolved: {
```
markgreenburg: And one more 😛

src/schemas/21-526EZ-allclaims/schema.js
```diff
@@ -594,78 +594,93 @@ const schema = {
...
+                type: 'array',
+                items: {
```
markgreenburg: I think this needs a `properties` property

src/schemas/21-526EZ-allclaims/schema.js
```diff
@@ -594,78 +594,93 @@ const schema = {
...
+                  name: {
+                    $ref: '#/definitions/fullName'
+                  },
+                  rank: {
+                    type: 'string'
+                  },
+                  injuryDeath: {
+                    type: 'string',
+                    enum: [
+                      'Killed in Action',
```
markgreenburg: We typically shy away from using nicely-formatted things in enum lists, because it's much more difficult to make content changes down the road, and any content changes you make will affect the api as well. Would recommend sticking to the categories like `['KILLED_ACTION', 'KILLED_NONBATTLE', 'WOUNDED_BATTLE'` etc and then managing the `enumNames` on the front end as needed.

src/schemas/21-526EZ-allclaims/schema.js
```diff
@@ -594,78 +594,93 @@ const schema = {
...
-              type: 'array',
-              items: {
-                type: 'object',
-                name: {
-                  $ref: '#/definitions/fullName'
-                },
-                address: {
-                  $ref: '#/definitions/address'
+              },
+              source: {
```
markgreenburg: one more singular for an array

src/schemas/21-526EZ-allclaims/schema.js
```diff
@@ -594,78 +594,93 @@ const schema = {
...
+                type: 'array',
+                items: {
```
markgreenburg: this looks like it's missing a `property` property as well

markgreenburg:
Oh also, sorry about the double-comments. I started leaving comments in the dist file but moved them over once I realized what I was looking at.

PR #307, opened 2018-12-18
All Claims - 781 - Changed PTSD incident source name type

dmccurdybah:
Updating the PTSD incident source.name from a `fullName` to a string. I'd thought that this was an individual, but was incorrect and this will have to be a single name field.

Added `description` as a string property to the PTSD incident individual.

### DIFF
--- a/dist/21-526EZ-ALLCLAIMS-schema.json
+++ b/dist/21-526EZ-ALLCLAIMS-schema.json
@@ -2583,7 +2583,7 @@
                   "type": "object",
                   "properties": {
                     "name": {
-                      "$ref": "#/definitions/fullName"
+                      "type": "string"
                     },
                     "address": {
                       "$ref": "#/definitions/address"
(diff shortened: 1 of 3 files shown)


dmccurdybah:
Hey @markgreenburg , thanks for the quick turnaround of the last PR, but this should be the last changes needed for 781. Could you review this when you get a chance?

dist/21-526EZ-ALLCLAIMS-schema.json
```diff
@@ -2583,7 +2586,7 @@
                   "type": "object",
                   "properties": {
                     "name": {
-                      "$ref": "#/definitions/fullName"
+                      "type": "string"
```
rianfowler: Does this need a unit test? I don't see a test validating `sources` here :https://github.com/department-of-veterans-affairs/vets-json-schema/blob/master/test/schemas/28-8832/schema.spec.js

dmccurdybah:
@rianfowler - Yes, I don't see anything for 526 at all.  Definitely something we'll need.  Can this get merged in whenever possible (we don't have the ability to merge this repo)

PR #308, opened 2018-12-21
All-Claims 781 Schema Update - personsInvolved name

dmccurdybah:
Updated fields:

personsInvolved.name - First name, middle name, and last name are not required for the PTSD 781 incident persons involved

ptsdIncident.incidentDescription - updated to match `form0781`

### DIFF
--- a/src/schemas/21-526EZ-allclaims/schema.js
+++ b/src/schemas/21-526EZ-allclaims/schema.js
@@ -168,7 +168,7 @@ const schema = {
       type: 'object',
       properties: {
         incidentDate: { $ref: '#/definitions/date' },
-        description: { type: 'string' },
+        incidentDescription: { type: 'string' },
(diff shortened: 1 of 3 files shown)


src/schemas/21-526EZ-allclaims/schema.js
```diff
@@ -646,7 +646,18 @@ const schema = {
                   type: 'object',
                   properties: {
                     name: {
-                      $ref: '#/definitions/fullName'
+                      type: 'object',
+                      properties: {
```
markgreenburg: The `fullName` definition also has `minLength` and `maxLength` properties for each of these keys, so this change will remove that restriction as well. Is this intended?

dmccurdybah: Correct, anything that exceeds that maximum length of the 21-0781 form field will overflow to an additional page

PR #311, opened 2018-12-26
0994 schema

imeecuison:
Schema for Vet Tec Form.

https://github.com/department-of-veterans-affairs/vets.gov-team/blob/bah-docs-sprint9/Products/Disability%20Compensation/BAH-526/engineering/VetTec/Form22-0994.pdf


src/schemas/22-0994/schema.js
```diff
@@ -0,0 +1,187 @@
...
+        '^[_A-Za-z0-9-]+(\\.[_A-Za-z0-9-]+)*@[A-Za-z0-9-]+(\\.[A-Za-z0-9]+)*(\\.[A-Za-z]{2,})$'
+    },
+    address: baseAddressDef,
+  },
+  properties: {
+    applicantFullName: {
+    $ref: "#/definitions/fullName"
+  },
+    applicantSocialSecurityNumber: {
+    $ref: "#/definitions/ssn"
```
dmccurdybah: Do we know if the applicant full name and social security number are going to be collected by the UI?  I know that the all-claims schema didn't actually include the veteran full name because it was actually added to the form data in `vets-api` from the user record instead of being sent from `vets-website`

If that was the case with the 0994, we need to make sure that the name isn't required.

src/schemas/22-0994/schema.js
```diff
@@ -0,0 +1,187 @@
...
+  "Media Application"
+];
+
+const salaries = [
+  "<$20,000",
+  "$20,001 - $35,000",
+  "$35,001 - $50,000",
+  "$50,001 - $75,000",
+  ">$75,000"
+];
```
dmccurdybah: I think all of these enum values can be defined in the property since they won't be used anywhere else since they're only used in one place.  Also, I think we should change these to something that is more "enum-like" (so something like `computerProgramming`, `dataProcessing`, or `lessThanTwenty` and then vets-api will map them to the form and the display values will be defined in the `vets-website` ui schema.

src/schemas/22-0994/schema.js
imeecuison: I think the name fields will be there but will be pre-populated with answers. As of right now, what I have done in ```bah-0994``` on vets-website is not populating though.

dmccurdybah:
You'll also need to increment the version number in `package.json`

imeecuison:
@department-of-veterans-affairs/code-reviewers-for-external-teams-group Could I get a code review on this at your earliest convenience. Thank you!

src/schemas/22-0994/schema.js
```diff
@@ -0,0 +1,177 @@
+import _ from "lodash/fp";
+import { countries } from "../../common/constants";
+import definitions from "../../common/definitions";
+import {
+  documentTypes526,
+  pciuCountries,
+  pciuStates
+} from "../../common/constants";
+
+const baseAddressDef = {
```
abhijitvk: @imeecuison - I don't see a need to define address separately in this schema. We can re-use the definition of `address` itself similar to the other education forms.

src/schemas/22-0994/schema.js
```diff
@@ -0,0 +1,177 @@
+import _ from "lodash/fp";
+import { countries } from "../../common/constants";
+import definitions from "../../common/definitions";
+import {
+  documentTypes526,
```
abhijitvk: Don't see `documentTypes526` being used anywhere

PR #315, opened 2019-01-08
Send DD-214 to ESR w/ health care application

lihanli:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/16104


lihanli:
how we've done it on other forms is, the file upload is done before the submission through a separate api, the api returns a uuid which the frontend saves and then sends along with the final submission

PR #317, opened 2019-01-09
Updated 8940 care dates

dmccurdybah:
`form8940` - Standardized `doctorProvidedCare` and `hospitalProvidedCare` property fields.

`address` is simplified to the standard address because all address fields are supported on the 8940 PDF.

`phoneNumber` was removed because it is not supported on the 8940 PDF.

Removed duplicate attachment enum values from `secondaryAttachment`, and `privateMedicalRecordAttachments`

Added `unemployabilityAttachments` and `employmentRequestAttachments`

### DIFF
--- a/dist/21-526EZ-ALLCLAIMS-schema.json
+++ b/dist/21-526EZ-ALLCLAIMS-schema.json
@@ -2891,11 +2891,8 @@
                   "address": {
                     "$ref": "#/definitions/address"
                   },
-                  "phoneNumber": {
-                    "$ref": "#/definitions/phone"
-                  },
                   "dates": {
-                    "$ref": "#/definitions/dateRange"
+                    "type": "string"
                   }
                 }
               }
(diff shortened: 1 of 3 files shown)


dmccurdybah:
@department-of-veterans-affairs/code-reviewers-for-external-teams-group - Could I get a code review on this at your earliest convenience? Thanks!

dist/21-526EZ-ALLCLAIMS-schema.json
```diff
@@ -3322,14 +3157,68 @@
...
+          "name": {
+            "type": "string"
+          },
+          "confirmationCode": {
+            "type": "string"
+          },
+          "attachmentId": {
+            "type": "string",
+            "enum": [
+              "L115"
```
rianfowler: Could probably use [`const`](https://json-schema.org/latest/json-schema-validation.html#rfc.section.6.1.3) here instead of an enum of one.

src/schemas/21-526EZ-allclaims/schema.js
```diff
@@ -1046,14 +1034,56 @@ const schema = {
...
-              'Decorations/Awards/Medal Citations',
               'Other Correspondence',
               'Buddy/Lay Statement'
             ]
           }
         }
-
+      }
+    },
+    unemployabilityAttachments: {
```
markgreenburg: @dmccurdybah @aaronjlech (@abhijitvk as FYI) sorry I didn't get a chance to review this PR before it was merged. I think the current way that `vets-api` parses the 526 form data to look for attachments is just to look at everything in the `attachments` key. So I think that any data submitted under an `unemployabilityAttachments` property will get ignored by `vets-api` and not actually attach to the form / pdf that's submitted to vbms. Can you provide some context on how / if this key is handled by `vets-api`?

Though not part of this PR, same goes for `employmentRequestAttachments`, `secondaryAttachment`, `completedFormAttachments`, and `privateMedicalRecordAttachments`. And any other attachment property that isn't called `attachments` 😛 

What probably needs to happen to enable this is:
1. Duplicate `attachments` property into `definitions`
2. Update the `attachments` property to `{ $ref: '#/definitions/vaTreatmentCenterAddress' }`
3. Delete all the other `___attachments` properties
4. In `vets-website`, update the page schemas for upload pages that need to have a restricted subset of selectable `attachmentId`s - they can use the `attachments` definition and modify from there.

Apologies if I've misunderstood how the api currently handles our attachments - if there's a plan to parse the submitted form data for all of these attachment keys, let me know.

PR #321, opened 2019-01-17
All claims fix capitalization mg

markgreenburg:
Updates all properties in the All Claims schema that have more than 2 adjacent capital letters, which would vets-api mangle them because of the camelCase to snakeCase to camelCase conversion process. E.g., isVAEmployee -> isVaEmployee.

Adds a `classificationCode` string property to satisfy https://github.com/department-of-veterans-affairs/vets.gov-team/issues/16355

### DIFF
--- a/dist/21-526EZ-ALLCLAIMS-schema.json
+++ b/dist/21-526EZ-ALLCLAIMS-schema.json
@@ -962,6 +962,9 @@
               "VA"
             ]
           },
+          "classificationCode": {
+            "type": "string"
+          },
           "primaryDescription": {
             "type": "string"
           },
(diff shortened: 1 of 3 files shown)


dist/21-526EZ-ALLCLAIMS-schema.json
```diff
@@ -2960,7 +2960,7 @@
                   "timeLostFromIllness": {
                     "type": "string"
                   },
-                  "mostEarningsInAMonth": {
+                  "mostEarningsInMonth": {
```
cvalarida: I think this one would be fine to keep as-is: `mostEarningsInAMonth` -> `most_earnings_in_a_month` -> `mostEarningsInAMonth`

PR #323, opened 2019-01-19
add option for non dd214 hca attachments

lihanli:
add option for non dd214 hca attachments


lihanli:
@erikphansen how does this option look for submitting non-dd214 attachments? for the backend submission we have a type code for dd214 attachments and another one for non-dd214 attachments

PR #330, opened 2019-02-04
Update 22-0994 schema

dmccurdybah:
`appliedForVAEducationBenefits` - Camel-cased name.

### DIFF
--- a/src/schemas/22-0994/schema.js
+++ b/src/schemas/22-0994/schema.js
@@ -53,7 +53,7 @@ const schema = {
     nightTimePhone: {
       $ref: "#/definitions/phone"
     },
-    appliedForVAEducationBenefits: {
+    appliedForVaEducationBenefits: {
       type: "boolean"
     },
     activeDuty: {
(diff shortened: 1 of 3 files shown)


dmccurdybah:
@department-of-veterans-affairs/frontend-review-group Could I get a code review on this at your earliest convenience. Thank you!

src/schemas/22-0994/schema.js
```diff
@@ -162,7 +162,7 @@ const schema = {
   required: [
     'privacyAgreementAccepted',
     'applicantFullName',
-    'appliedForVAEducationBenefits',
+    'appliedForVaEducationBenefits',
```
jbalboni: Do you need to add this in here? It won't make it required that this value is true, it just requires that the value is defined.

dmccurdybah: Good question!

I know that we do want an answer either way because It's used by the regional office to know whether or not they need to find an existing 22-1990 for the applicant.

dmccurdybah:
@jbalboni could you merge this when you get a chance?

PR #335, opened 2019-02-19
Make YAAS: Yet Another Address Schema

cvalarida:
The incident sources address doesn't populate the `addressLine3` to the PDF, so rather than having a veteran fill it in with some information only to be completely ignored, we'll just not show them the field.

I couldn't just hide the field on `vets-website` because throwing in a `'ui:options': { hideIf: () => true }` doesn't work on the first render, and the address schema was a reference to the definition (and I thus didn't have access to remove it or apply `ui:hidden`).


cvalarida:
Thanks. Hopefully this is the last for this change. :crossed_fingers:

PR #338, opened 2019-03-18
Add caret to homelessness contact name pattern

markgreenburg:
Adds caret in homelessness POC regex to match beginning of string, so that validations actually applied.


src/schemas/21-526EZ-allclaims/schema.js
```diff
@@ -100,16 +100,16 @@ const form0781AddressDef = (addressSchema => {
     'addressLine2',
     'addressLine3',
     'postalCode',
-    'zipCode',
+    'zipCode'
```
markgreenburg: Sorry, format-on-save did some harmless stuff 🤷‍♂️

PR #342, opened 2019-04-08
alter regex to eliminate warnings that appear in `vets-api` when running specs

jholton:
These regex as written cause the vets-api to throw warnings when running specs.
Rewriting them this way eliminates the warnings.

[StackOverflow reference](https://stackoverflow.com/questions/24843335/rails-regex-warning-character-class-has-without-escape/24843354)


jholton:
Previous PRs to fix these warnings:  [1](https://github.com/department-of-veterans-affairs/vets-json-schema/pull/224), [2](https://github.com/department-of-veterans-affairs/vets-json-schema/pull/284)

PR #343, opened 2019-04-09
Feedback schema key rename

lihanli:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/17611
change feedback schema keys to be all camelcase


rianfowler:
We'll need to coordinate rollout of this change with the front end. Have looped in @rtluu and @mattpointzxer0

PR #344, opened 2019-04-15
Add 526 boolean for terminallyIll

hinzed1127:
Corresponding schema change for department-of-veterans-affairs/vets-website#9930

### DIFF
--- a/dist/21-526EZ-ALLCLAIMS-schema.json
+++ b/dist/21-526EZ-ALLCLAIMS-schema.json
@@ -1016,7 +1016,7 @@
         }
       }
     },
-    "terminallyIll": {
+    "isTerminallyIll": {
       "type": "boolean"
     },
     "vaTreatmentFacilities": {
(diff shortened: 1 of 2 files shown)


dist/21-526EZ-ALLCLAIMS-schema.json
```diff
@@ -1016,6 +1016,9 @@
         }
       }
     },
+    "terminallyIll": {
```
SMLuthi: The BE is looking for `isTerminallyIll` in the form.

src/schemas/21-526EZ-allclaims/schema.js
```diff
@@ -488,6 +488,9 @@ const schema = {
         }
       }
     },
+    terminallyIll: {
```
SMLuthi: Same here

dist/21-526EZ-ALLCLAIMS-schema.json
hinzed1127: sounds good, will update

SMLuthi: Thanks!

PR #350, opened 2019-05-29
18656: STEM: Update 1995 Education Benefit Page

elder-joseph:
Addition of 'fryScholarship' as benefit type to the 1995 Schema

### DIFF
--- a/dist/22-1995-schema.json
+++ b/dist/22-1995-schema.json
@@ -395,6 +395,7 @@
       "type": "string",
       "enum": [
         "chapter33",
+        "fryScholarship",
         "chapter30",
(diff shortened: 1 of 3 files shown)


dmccurdybah:
Do we want to do this?  Isn't the purpose of handling this in the 1995 transformer because "fryScholarship" shouldn't be in the schema?

elder-joseph:
@dmccurdybah That would greatly simplify the PRs for this story if we did not want this value in the schema. The benefit selection options for the 1995 are currently populated from the values of that list from the schema. We could perhaps specifically add the value to that list explicitly if we wish to avoid adding it to the schema.

PR #352, opened 2019-07-25
[526EZ All-Claims] Add character limit to otherAtRiskHousing

ncksllvn:
This PR should resolve the prod error below, but this is my first PR to vets-json-schema so 🤞 🙂 

```
form526.veteran.homelessness.homelessnessRisk.otherLivingSituation.Size / “size must be between 1 and 500”
```

Ticket - https://github.com/department-of-veterans-affairs/vets.gov-team/issues/19329

### DIFF
--- a/src/schemas/21-526EZ-allclaims/schema.js
+++ b/src/schemas/21-526EZ-allclaims/schema.js
@@ -473,7 +473,8 @@ const schema = {
       enum: ['losingHousing', 'leavingShelter', 'other']
     },
     otherAtRiskHousing: {
-      type: 'string'
+      type: 'string',
+      maxLength: 500,
     },
     homelessnessContact: {
       type: 'object',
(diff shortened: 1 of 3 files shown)


src/schemas/21-526EZ-allclaims/schema.js
```diff
@@ -473,7 +473,8 @@ const schema = {
       enum: ['losingHousing', 'leavingShelter', 'other']
     },
     otherAtRiskHousing: {
-      type: 'string'
+      type: 'string',
+      maxLength: 500,
```
ncksllvn: I'm assuming that this is the property translated to the FE, whereas the prod error will show the untranslated version. I think this is done [here](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/evss/disability_compensation_form/data_translation_all_claim.rb#L359).

ncksllvn:
Thanks @lihanli! @cvalarida Can you also check this out when you have moment?

ncksllvn:
Thanks @cvalarida, opened a vets-website pull request [here](https://github.com/department-of-veterans-affairs/vets-website/pull/10498) following your direction!

PR #356, opened 2019-09-05
Rename and update 1995-STEM

dmccurdybah:
Updated 22-1995-STEM to match the current 1995 schema, and renamed it to 22-1995S.

The current 22-1995-STEM is a placeholder that isn't currently being used. The intention is that it will be a copy of the 1995 schema that could potentially change in the future.


dmccurdybah:
@department-of-veterans-affairs/frontend-review-group Could I get a code review on this at your earliest convenience. Thank you!

PR #365, opened 2020-01-09
Add ESLint, and Prettier formatting

LinnJS:
- Added ESLint and Prettier formatting for ease of workflow when working with this repo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,2 @@
 node_modules
 /.project
-
-# WebStorm indexing
-.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 node_modules
 /.project
+
+# WebStorm indexing
+.idea
--- a/src/common/constants.js
+++ b/src/common/constants.js
@@ -1,7 +1,7 @@
 import _ from 'lodash';
 import vaMedicalFacilities from './va-medical-facilities';
 
-const countries = [
+export const countries = [
   { value: 'USA', label: 'United States' },
   { value: 'AFG', label: 'Afghanistan' },
   { value: 'ALB', label: 'Albania' },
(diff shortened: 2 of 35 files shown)


LinnJS:
@Mottie I realized that I had to setup prettier separately, the rules were conflicting a bit. But once I added a `.pretierrc` it fixed it right away. All formatting should be the same across all files now.

src/common/constants.js
```diff
@@ -2800,252 +2883,243 @@ const days = {
...
-  'Daughter',
-  'Son',
-  'Stepson',
-  'Stepdaughter',
-  'Father',
-  'Mother',
-  'Spouse',
-  'Other'
-];
+const dependentRelationships = ['Daughter', 'Son', 'Stepson', 'Stepdaughter', 'Father', 'Mother', 'Spouse', 'Other'];
```
Mottie: Not sure why this array is being moved into a single line. I don't think it should be modified.

Is there a max character limit per line set (I forget what the lint setting is named)?

LinnJS: Yes it looks like that was changed to one line because it did not hit the char limit of 120 chars. I can make that shorter, or look for that rule and disable it, or leave it be. I don't really have an opinion to be honest I think this is just Prettier Standard.

.gitignore
```diff
@@ -1,2 +1,5 @@
 node_modules
 /.project
+yarn-error.log
+# WebStorm indexing
+.idea
```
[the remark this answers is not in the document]
LinnJS: @cvalarida I removed it from our .gitignore, had no idea this was a thing, thanks for sharing!

src/common/constants.js
Mottie: Hmm, I thought that was set to 80 everywhere else.

PR #367, opened 2020-01-10
4494 686c schema updates

micahchiang:
[Original ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/4494)

## Summary of Issue
- The current workflow in the the 686 supports two of the 8 possible workflows the form can be used for. This ticket is to add fields that are missing to the schema file as well as to create a generic location schema that can be used for multiple location inputs on the form, including the location(s) of marriage(s), so that the existing workflows can map 1 to 1 with the paper form of the 686.
- This PR addresses these issues:
    * Amending how a user enters their location of marriage
    * Adding new properties to the schema that weren't there before.
    * Updating the names for some properties to more closely align with their purpose.

## Considerations
- This is the first of potentially several PRs that will be made to update the 686 schema. As more workflows are added, additional properties will need to be included.

### DIFF
--- a/src/common/definitions.js
+++ b/src/common/definitions.js
@@ -126,6 +126,13 @@ const ssn = {
   pattern: '^[0-9]{9}$'
 };
 
+// Historically a veteran's service number has been between 5 and 8 digits, 
+// with the possibility of an alphabetical prefix of up to two capital letters.
+const veteranServiceNumber = {
+  type: 'string',
+  pattern: '^[A-Z]{0,2}\\d{5,8}$'
+}
+
 // The last four digits (or serial number) must be a number from 0001 to 9999
 // https://www.ssa.gov/history/ssn/geocard.html
 const ssnLastFour = {
--- a/src/common/definitions.js
+++ b/src/common/definitions.js
@@ -124,12 +124,12 @@ const ssn = {
   pattern: '^[0-9]{9}$',
 };
 
-// Historically a veteran's service number has been between 5 and 8 digits, 
+// Historically a veteran's service number has been between 5 and 8 digits,
 // with the possibility of an alphabetical prefix of up to two capital letters.
 const veteranServiceNumber = {
   type: 'string',
-  pattern: '^[A-Z]{0,2}\\d{5,8}$'
-}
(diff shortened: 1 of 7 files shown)


src/common/definitions.js
```diff
@@ -126,6 +126,13 @@ const ssn = {
   pattern: '^[0-9]{9}$'
 };
 
+// Historically a veteran's service number has been between 5 and 8 digits, 
+// with the possibility of an alphabetical prefix of up to two capital letters.
+const veteranServiceNumber = {
```
micahchiang: I elected to add this to the common definitions because it seems there's a possibility it may be used elsewhere. After speaking with folks at the DSVA, and doing some research, this pattern was chosen because historically, service numbers can have between 5 and 8 numbers, prefixed by up to two capital letters.

src/schemas/21-686C/schema.js
```diff
@@ -74,12 +74,19 @@ const commonAddressFields = {
...
-  required: ['dateOfMarriage', 'locationOfMarriage', 'spouseFullName'],
+  required: ['dateOfMarriage', 'locationOfMarriage', 'spouseFullName', 'marriageType'],
   properties: {
     dateOfMarriage: schemaHelpers.getDefinition('date'),
     locationOfMarriage: {
-      $ref: '#/definitions/location'
+      $ref: '#/definitions/genericLocation'
+    },
+    marriageType: {
+      type: 'string'
```
micahchiang: Similar to the pensions form. This is a field required on the paper 686, and needed to be added here.

src/schemas/21-686C/schema.js
```diff
@@ -474,11 +500,38 @@ let schema = {
...
+                  reasonMarriageEnded: {
+                    type: 'string',
+                    enum: [
+                      'Other',
+                    ]
+                  },
+                  explainSeparation: {
+                    type: 'string',
+                    maxLength: 500,
+                    pattern: textAndNumbersRegex
```
micahchiang: As it pertains to a child's previous marriage the form only asks for the date the marriage ended and the reason, as opposed to the date they were married. These fields were added to address this.

PR #381, opened 2020-02-12
Update 10-10CG schema to use vaMedicalFacilities enum list for vetera…

kevinmirc:
…n.plannedClinic.

**Note**: This form is not being consumed by vets-api yet, so no need for data migration even though this is considered a breaking change.


src/examples/10-10CG/example.js
```diff
@@ -9,7 +9,7 @@ const example = {
     alternativePhoneNumber: '8887775544',
     email: 'veteranEmail@email.com',
     vaEnrolled: true,
-    plannedClinic: 'My Clinic',
+    plannedClinic: '635GF',
```
LinnJS: Cool so I guess we are going to use facility code? Love that if thats the case so if the name changes we need not to worry.

PR #383, opened 2020-02-21
MDOT-schema.json Update

mr0sari0:
## Description
Updating `MDOT-schema.json` to include necessary properties so we can use UI schemas to display error messages.

## Testing done
Used ESLint for validation

## Definition of done
- [ ] Events are logged appropriately
- [ ] Documentation has been updated, if applicable
- [ ] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)
- [ ] No sensitive information (i.e. PII/credentials/internal URLs/etc.) is captured in logging, hardcoded, or specs


mr0sari0:
@LinnJS and @camerontesterman Hey guys, whenever you get a chance, would appreciate a review. Thanks!

src/schemas/MDOT/schema.js
```diff
@@ -1,34 +1,217 @@
...
+    city: {
+      type: "string",
+      minLength: 1,
+      maxLength: 51
+    },
+    state: {
+      title: "State",
+      type: "string",
+      maxLength: 51,
+      enum: [
```
Mottie: This list of states should be coming from `src/common/constants.js`

If you only need the 50 states & DC, then use `states50AndDC`; if you want to include overseas bases & territories, then use `states.USA`.

PR #395, opened 2020-04-09
[Tech debt] Split up constants

Mottie:
## Description

While attempting to consolidate all the data, I ended up doing the following:

- Moved all address related constants into a separate `address.js` file. 
- Cross-referenced our list of countries with [this site](https://worldpopulationreview.com/countries/how-many-countries-are-there/) which pulls its data from https://github.com/mledoze/countries
- Added `militaryCities` & `militaryStates`, which is of common use with the new overseas military checkbox design
- Added `addressType` & `addressPou`, which is needed in schemas.

## Testing Done

Local unit tests


Mottie:
Note: If you use `gitk` during the review of this PR, it's a little bit easier to see that the only changes made to the `countries` variable are the comments and the commented out block at the end.

PR #404, opened 2020-04-17
Remove vic

bkjohnson:
We should no longer need the `vic` schema, and having it is causing a test to fail that compares the number of schemas. https://github.com/department-of-veterans-affairs/vets-website/pull/12254#issuecomment-614926892


bkjohnson:
@department-of-veterans-affairs/backend-review-group I took a look through vets-api and it didn't look like any of the places where `VetsJsonSchema` is being used were vic-related.  Can you confirm?

PR #405, opened 2020-04-17
bradl3yC - 7099 - Update schema for form 40-10007

bradl3yC:
Update the schema to reflect the changes requested in va.gov/team 7099

### DIFF
--- a/src/schemas/40-10007/schema.js
+++ b/src/schemas/40-10007/schema.js
@@ -237,7 +237,8 @@ const schema = {
             },
             ethnicity: {
               type: 'string',
-              enum: ['American Indian or Alaska Native', 'Black or African American', 'Hispanic or Latino', 'Native Hawaiian or Pacific Islander', 'White']
+              enumNames: ['American Indian or Alaska Native', 'Black or African American', 'Hispanic or Latino', 'Native Hawaiian or Pacific Islander', 'White'],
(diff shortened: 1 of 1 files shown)


src/schemas/40-10007/schema.js
```diff
@@ -234,6 +235,10 @@ const schema = {
               type: 'string',
               enum: ['Female', 'Male'],
             },
+            ethnicity: {
+              type: 'string',
+              enum: ['American Indian or Alaska Native', 'Black or African American', 'Hispanic or Latino', 'Native Hawaiian or Pacific Islander', 'White']
```
mr0sari0, line 240: `enums` should be camelCased. If you want, create a `enumNames` property that hold what you want the user to see.

LinnJS: Oh yeah good catch, and then use labels to make them look like this

src/schemas/40-10007/schema.js
```diff
@@ -234,6 +235,11 @@ const schema = {
               type: 'string',
               enum: ['Female', 'Male'],
             },
+            ethnicity: {
+              type: 'string',
+              enumNames: ['American Indian or Alaska Native', 'Black or African American', 'Hispanic or Latino', 'Native Hawaiian or Pacific Islander', 'White'],
```
LinnJS, line 240: You dont need enumName, need to put those in the `ui:schema`

PR #408, opened 2020-04-20
Revert "Remove vic"

bradl3yC:
Reverts department-of-veterans-affairs/vets-json-schema#404


bradl3yC:
This is currently a breaking change in vets-website as the VIC json schema is still in use.

PR #409, opened 2020-04-20
Fix states constants

Mottie:
When I split out the address data from the `constants.json` file (https://github.com/department-of-veterans-affairs/vets-json-schema/pull/395), I also removed lodash.

In the process, I changed

```js
const usaStates = _.map(states.USA, stateData => {
  return stateData.value;
});
```

to

```js
const usaStates = Object.values(states.USA);
```

which was a silly mistake.

This PR changes that code to:

```js
const usaStates = states.USA.map(state => state.value);
```

### DIFF
--- a/dist/constants.json
+++ b/dist/constants.json
@@ -1,4 +1,741 @@
 {
+  "addressPou": {
+    "correspondence": "CORRESPONDENCE",
+    "residence": "RESIDENCE/CHOICE"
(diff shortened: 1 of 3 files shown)


Mottie:
Sorry, this looks like a HUGE change, but all I did was sort the export, so now the data has been re-ordered in the dist.

PR #410, opened 2020-04-20
Update README.md

kevinmirc:


README.md
```diff
@@ -13,3 +13,16 @@ node v6.11.1
...
+- Update the "version" property in `package.json` with the new version. Please follow [Semantic Versioning](https://semver.org/#summary) practices.
+  - If implementing a major version update, add a link in your vets-json-schema PR that references the vets-api, or vets-website, PR that addresses the breaking changes.
+  - Breaking changes include:
+    - Removing a property from a schema
+    - Adding a property to a schema that has `additionalProperties` set to false
+    - Making a property required that was not previously required on that schema
+    - Changing the `type` on a schema's property
+    - Removing values in the `enum` key of a schema's property
+    - ect.
+- Run `npm update` or `yarn update` to update `package-lock.json`.
```
cvalarida, line 27: Will `yarn update` update the `package-lock.json`, or will it only update the `yarn.lock`?

kevinmirc: I took a guess at what the equivalent yarn command was. But yes `yarn upgrade`* will only update yarn.lock. I just removed that instruction since npm does the trick.

PR #413, opened 2020-04-20
add is_military_base and prescribed_date to mdot schema

camerontesterman:


src/schemas/MDOT/schema.js
```diff
@@ -1,8 +1,14 @@
...
 import schemaHelpers from '../../common/schema-helpers';
+import _ from 'lodash';
 
 const { fullName, ssnLastFour, email, gender, date, address } = commonDefinitions;
 
+const addressWithIsMilitaryBase = _.cloneDeep(address);
+addressWithIsMilitaryBase.properties.isMilitaryBase = {
+  type: 'boolean'
+};
+
```
bradl3yC, line 11: I think we can do something more es6-ish with this
```
const addressWithIsMilitaryBase = {
  ...address,
  isMilitaryBase: {
    type: 'boolean'
  }
}
```

bradl3yC: this will alleviate the need to import lodash

PR #432, opened 2020-05-20
9206 - Update schema properties for 686

micahchiang:
## Description
[Original Ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/9206)

This pull request updates the schema properties for some fields in the 686. This was needed to better align with accepted values in BGS. The fields for the report a divorce workflow are also being renamed to `reasonMarriageEnded` and `explanationOfOther` to provide more semantic context to their purpose.

### DIFF
--- a/src/schemas/686c-674/schema.js
+++ b/src/schemas/686c-674/schema.js
@@ -382,8 +382,8 @@ const schema = {
                   },
                   reasonMarriageEnded: {
                     type: 'string',
-                    enum: ['DIVORCE', 'DEATH', 'ANNULMENT', 'OTHER'],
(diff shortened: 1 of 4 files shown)


src/schemas/686c-674/schema.js
```diff
@@ -478,12 +478,13 @@ const schema = {
         location: {
           $ref: '#/definitions/genericLocation',
         },
-        isMarriageAnnulledOrVoid: {
-          $ref: '#/definitions/genericTrueFalse',
-        },
-        explanationOfAnnullmentOrVoid: {
+        reasonMarriageEnded: {
```
micahchiang, line 481: Renamed this property and the one below it, since we're no longer explicitly asking if the marriage was annulled or declared void.

PR #441, opened 2020-06-22
MDOT Schema Update/ Add addressBuilder Function

mr0sari0:
# Description
This PR deals with property and requirement updates on MDOT schema.

Part of the challenge was thinking about how to use the common address definition with our data. 

The current implementation of `address` does not support full country names so I decided to create `addressBuilder`. This function accepts a parameter that will create a schema for that case of there being full country names in your data.

For email validation, I suggest we use `format: 'email'` since it takes care of most of the validation for us.

## Acceptance Criteria

- Required properties have been updated
- Country and state properties have been added
- Email validation has been updated to accept special characters before the @ sign


src/common/address.js
```diff
@@ -384,6 +386,19 @@ const states = {
...
+  statesCopy.Canada = statesCopy.CAN;
+  statesCopy['United States'] = statesCopy.USA;
+  statesCopy.Mexico = statesCopy.MEX;
+  delete statesCopy.CAN;
+  delete statesCopy.USA;
+  delete statesCopy.MEX;
+  return statesCopy;
+};
+
+const statesWithFullCountryNames = buildStatesWithFullCountryNames();
```
Mottie, line 400: It seems to me that this could be solved on the front-end with some creative labeling... adding a full list of `statesWithFullCountryNames` duplicates a lot of content in the `constants.json` making it bigger than it needs to be.

src/common/definitions.js
```diff
@@ -125,6 +125,92 @@ const address = (() => {
...
+        type: 'string',
+      },
+      street: {
+        type: 'string',
+        minLength: 1,
+        maxLength: 50,
+      },
+      street2: {
+        type: 'string',
+        minLength: 1,
```
Mottie, line 189: Including a `minLength` makes this field required, I believe.

src/common/definitions.js
```diff
@@ -125,6 +125,92 @@ const address = (() => {
...
+    useCountryFullName ? constants.statesWithFullCountryNames : constants.states,
+    (value, key) => ({
+      properties: {
+        country: {
+          type: 'string',
+          enum: [key],
+        },
+        state: {
+          type: 'string',
+          enum: value.map(x => x.value),
```
Mottie, line 143: It seems like not including `enumNames` would only show lower case full-state names for Canada & Mexico

src/common/address.js
mr0sari0: The reason for this function and `addressBuilder` is because our backend is providing us full country names and the current `address` common definition is looking for the abbreviations of the countries:

```
const address = (() => {
  // eslint-disable-next-line import/no-named-as-default-member
  const countries = constants.countries.map(object => object.value);
...
})();```

File: src/common/definitions.js (hunk 3 above)
Comment (mr0sari0) [reply src/common/definitions.js:143]: This was copied from the original `address` definition which is being used on most schemas.

```
const address = (() => {
  ...
  const countryStateProperties = _.map(constants.states, (value, key) => ({
    properties: {
      country: {
        type: 'string',
        enum: [key],
      },
      state: {
        type: 'string',
        enum: value.map(x => x.value),
      },
      postalCode: {
        type: 'string',
        maxLength: 10,
      },
    },
  }));
 ...
})();
```

src/common/definitions.js
mr0sari0: Just removed it.

src/common/address.js
Mottie: That's fine, but it shouldn't need to be added to the `constants.json`

mr0sari0: I was following suit with how `states` was being imported and exported but I can provide a direct reference from `address.js`.

mr0sari0: Just pushed up a refactor.

lihanli:
why is the full country name necessary for the schema? if the mdot API requires the full country name can we convert it on the backend from the iso3 code before submission and continue to use the common address definition?

mr0sari0:
> why is the full country name necessary for the schema? if the mdot API requires the full country name can we convert it on the backend from the iso3 code before submission and continue to use the common address definition?

From what I understand, the DLC is sending us the full country name. If they were sending the abbreviated forms, I wouldn't need this `addressBuilder` function or the `statesWithFullCountryNames` and I could just display the full names using the country labels in `vets-website`.

lihanli:
> > why is the full country name necessary for the schema? if the mdot API requires the full country name can we convert it on the backend from the iso3 code before submission and continue to use the common address definition?
> 
> From what I understand, the DLC is sending us the full country name. If they were sending the abbreviated forms, I wouldn't need this `addressBuilder` function or the `statesWithFullCountryNames` and I could just display the full names using the country labels in `vets-website`.

OK that makes sense, but do we know if the country names the DLC uses are the same as the country names we have? for example america could be called United States or it could be called United States of America. i think we should either get the list of country names from them or ideally make them use the iso3 code instead of the full country name

mr0sari0:
> > > why is the full country name necessary for the schema? if the mdot API requires the full country name can we convert it on the backend from the iso3 code before submission and continue to use the common address definition?
> > 
> > 
> > From what I understand, the DLC is sending us the full country name. If they were sending the abbreviated forms, I wouldn't need this `addressBuilder` function or the `statesWithFullCountryNames` and I could just display the full names using the country labels in `vets-website`.
> 
> OK that makes sense, but do we know if the country names the DLC uses are the same as the country names we have? for example america could be called United States or it could be called United States of America. i think we should either get the list of country names from them or ideally make them use the iso3 code instead of the full country name

That would be a @camerontesterman question.

PR #454, opened 2020-07-16
Fix validation error on postal code

LinnJS:
# New schema
- Combine patterns into one pattern for postal code
- Added test cases for postalCode validation

## Screenshot
image


src/common/definitions.js
```diff
@@ -53,14 +53,7 @@ const rejectOnlyWhitespace = {
...
   type: 'string',
-  anyOf: [
-    {
-      pattern: '^\\d{5}$',
-    },
-    {
-      pattern: '^\\d{5}-\\d{4}$',
-    },
-  ],
+  pattern: '^\\d{5}(-\\d{4})?$',
```
LinnJS, line 56: Removed `anyOf` since zipCode frontend tooling does not account for this use case of `react-JSON-schema`

kevinmirc:
@LinnJS I think you still need to run the build and commit the changes to `dist`.

src/common/definitions.js
kevinmirc: I see another "postalCode" regex that is very similar to this already the codebase. Maybe we should use the same pattern to be consistent?

`^\\d{5}(?:([-\\s]?)\\d{4})?$`

PR #458, opened 2020-07-21
11338 field updates

cohnjesse:
# Update to existing schema

[Original ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/11338)

This PR is to add a field for the checkbox to switch between state and country fields as well as to update the state field to be an enum. This PR also updates the unit tests accordingly.


src/schemas/686c-674/schema.js
```diff
@@ -20,15 +20,24 @@ const schema = {
...
-        state: {
+        isOutsideUS: {
+          type: 'boolean',
+          default: false,
+        },
+        country: {
           type: 'string',
           maxLength: 30,
-          pattern: textOnlyPattern,
+          pattern: '^(?!\\s)(?!.*?\\s{2,})[^<>%$#@!^&*0-9]+$',
```
micahchiang, line 30: I've seen this pattern before, but am still a little unclear on what exactly it checks for O.o

PR #472, opened 2020-08-06
Add json schema for IRIS Ask a Question

shefalinayak:
# New schema

department-of-veterans-affairs/orchid#17

This schema is for the new IRIS Ask a Question form (replacing https://iris.custhelp.va.gov/app/ask) that will live on VA.gov.

vets-website PR to be created
vets-api PR to be created

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "6.6.3",
+  "version": "6.7.3",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/department-of-veterans-affairs/vets-json-schema.git"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "6.7.3",
+  "version": "6.7.0",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/department-of-veterans-affairs/vets-json-schema.git"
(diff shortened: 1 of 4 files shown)


stephenBarrs:
Hi! Just FYI in case you do not know, when adding a new Schema to `vets-json-schema` you have to add a test to `vets-website` or to an exclusion list or the `vets-website` tests will break (I found this out the difficult way!).  Here is the exclusion list: https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/platform/forms/tests/forms.unit.spec.js#L71

shefalinayak:
> Hi! Just FYI in case you do not know, when adding a new Schema to `vets-json-schema` you have to add a test to `vets-website` or to an exclusion list or the `vets-website` tests will break (I found this out the difficult way!). Here is the exclusion list: https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/platform/forms/tests/forms.unit.spec.js#L71

@stephenBarrs I did not know, thank you for the info. Created a [vets-website PR](https://github.com/department-of-veterans-affairs/vets-website/pull/13808) to add the form to the exclusion list.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "6.6.3",
+  "version": "6.7.3",
```
f1337, line 3:
```suggestion
  "version": "6.7.0",
```

Per [semver](https://semver.org), shouldn't this be `6.7.0`?

shefalinayak: Right, sorry we missed that!

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "6.6.3",
+  "version": "6.7.0",
```
f1337, line 3: Please double-check the version in `master` before merging? There are several PRs open right now. Thank you!!

PR #473, opened 2020-08-07
Update README.md

cvalarida:
Updates the README per https://github.com/department-of-veterans-affairs/va.gov-team/issues/12023.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,20 @@
 # vets-json-schema [![Build Status](https://travis-ci.org/department-of-veterans-affairs/vets-json-schema.svg?branch=master)](https://travis-ci.org/department-of-veterans-affairs/vets-json-schema)
 
+## Purpose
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,15 +1,29 @@
...
+### Workflow
+1. Clone the repo
+1. Make a branch for your changes
+1. Make the changes needed for your form
+1. Update the version number in `package.json`
+1. Submit a PR
+1. Once that PR is merged
+    1. Create a PR in `vets-website` to update the `vets-json-schema` dependency
+        - This can be done with `yarn update:schema`
+    1. Create a PR in `vets-api` to update the dependency
```
U-DON, line 17: Not that this ends up mattering since it's corrected in the readable view, but these are all 1.

cvalarida: Yeah, I read somewhere that order lists should all be `1` because it makes modifying them later easier. :man_shrugging:

PR #476, opened 2020-08-11
Feature/covid research update

stephenBarrs:
# Update schema
Update to COVID-VACCINE-SCHEMA to remove view: fields and update required fields

department-of-veterans-affairs/vets-website#0000  Coming Soon
department-of-veterans-affairs/vets-api#0000 Coming Soon


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "6.6.3",
+  "version": "6.7.4",
```
f1337, line 3:
```suggestion
  "version": "6.7.0",
```

Per [semver](https://semver.org), shouldn't this be `6.7.0`?

stephenBarrs: @f1337 I based the patch increment on the PR in the Queue prior to this PR per this [Slack](https://dsva.slack.com/archives/CBU0KDSB1/p1597149227305700). I think their minor increment should be `6.7.0` making this patch increment `6.7.1`.   Getting the version of these correct with multiple PRs that might be merged at different times seems difficult.   Thanks!

stephenBarrs: Changed version to 6.7.1 based on 6.7.0 set in #472

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "6.6.3",
+  "version": "6.7.1",
```
f1337, line 3: If this PR merges first, I'd take 6.7.0, then let the other PR author know they need to update to 6.7.1 before they merge.

PR #478, opened 2020-08-12
version check script for ci

f1337:
See https://app.zenhub.com/workspaces/vsp-5cedc9cce6e3335dc5a49fc4/issues/department-of-veterans-affairs/va.gov-team/12412 for details.

### DIFF
--- a/bin/version_check
+++ b/bin/version_check
@@ -1,0 +1,11 @@
+#! /usr/bin/env bash
+
+# USAGE: ./bin/version_check || circleci step halt
+
+changes=`git show -m HEAD --name-only --pretty=""`
+json_changes=`echo $changes | egrep -q json && echo "1"`
+schema_changes=`echo $changes | egrep -q 'src/schemas' && echo "1"`
+
+# return 1 (bash error) if there are json or schema changes
+if [[ $json_changes = 1 ]]; then exit 1; fi
+if [[ $schema_changes = 1 ]]; then exit 1; fi
--- a/dist/10-10CG-example.json
+++ b/dist/10-10CG-example.json
@@ -1,7 +1,7 @@
 {
   "veteran": {
     "fullName": {
-      "first": "Jane Doe",
+      "first": "Jan Doe",
       "last": "Doe"
     },
     "ssnOrTin": "789787893",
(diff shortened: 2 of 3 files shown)


dist/10-10CG-example.json
```diff
@@ -1,7 +1,7 @@
 {
   "veteran": {
     "fullName": {
-      "first": "Jane Doe",
+      "first": "Jan Doe",
```
f1337, line 4: This is a test change, to trigger a CI failure. It will not be merged.

omgitsbillryan:
So any changes to `.json` files or any files in `/src/schema` will result in a circleci failure - am I reading this properly?

I'm a `vets-json-schema` rookie, but looking at the past few commits to this repo it seems like normal workflow to change these files. I think I'm missing something.

bin/version_check
```diff
@@ -0,0 +1,11 @@
+#! /usr/bin/env bash
+
+# USAGE: ./bin/version_check || circleci step halt
+
+changes=`git show -m HEAD --name-only --pretty=""`
```
omgitsbillryan, line 5: I need to be reminded - what does `git show` show? Is it only the last commit? What if a PR has multiple commits?

f1337: > So any changes to `.json` files or any files in `/src/schema` will result in a circleci failure - am I reading this properly?

Right, this is just the first step, making sure the bash script can fail CircleCI.

Next step is to only fail if there are those changes and package.json isn't also changes.

Then I want to actually check that the version string of package.json is updated.

The stretch goal would be to check that the version string is *incremented* in package.json. But I think the tool is still useful if we stop at the immediate previous step.

f1337:
> So any changes to `.json` files or any files in `/src/schema` will result in a circleci failure - am I reading this properly?

Right, this is just the first step, making sure the bash script can fail CircleCI.

Next step is to only fail if there are those changes and package.json isn't also changes.

Then I want to actually check that the version string of package.json is updated.

The stretch goal would be to check that the version string is *incremented* in package.json. But I think the tool is still useful if we stop at the immediate previous step.

bin/version_check
f1337: Yeah, it'll be `git diff origin/master` in the next iteration.

PR #480, opened 2020-08-17
Add Danger.js

f1337:
# Description

Adds [Danger.js](https://danger.systems/js/), so that we can introduce smarter CI warnings. The first intended use is to [warn when JSON or schema files change without bumping the version](https://app.zenhub.com/workspaces/vsp-5cedc9cce6e3335dc5a49fc4/issues/department-of-veterans-affairs/va.gov-team/12412).

- [x] Merge #481 
- [x] Change the base for this PR to `master`

# Related Issues

@omgitsbillryan opened an issue re: [creating a danger-specific user](https://github.com/department-of-veterans-affairs/va.gov-team/issues/7551) with **much less access** than the current vfs bot requires. TL;DR: Danger (.js or .rb) needs the absolute minimum access possible, while our vfs bot has broad privileges. We should fix this because securitah.

# New schema

**NONE**

_N/A_

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,6 +10,7 @@ jobs:
       - node/with-cache:
           steps:
             - run: npm install
+            - run: yarn danger ci
             - run: npm test
 workflows:
     build-and-test:
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -1,0 +1,1 @@
+import { danger, fail } from "danger";
--- a/package-lock.json
+++ b/package-lock.json
@@ -900,6 +900,16 @@
         "@babel/helper-plugin-utils": "7.10.4"
       }
     },
+    "@babel/polyfill": {
+      "version": "7.10.4",
+      "resolved": "https://registry.npmjs.org/@babel/polyfill/-/polyfill-7.10.4.tgz",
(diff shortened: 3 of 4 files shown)


.nvmrc
```diff
@@ -1 +1 @@
-6.4.0
+10.13.0
```
omgitsbillryan, line 1: I think there's a reason this has remained at 6, but I don't know what that reason is.

omgitsbillryan:
Looks like its on the right track to me. I think you may want to split this work into 3 separate PRs:

1. Changing the .nvmrc file
2. Add danger to CI where no checks are performed, it just auto-passes
3. Add your schema version checks where danger may now fail CI builds

.nvmrc
f1337: Oddly, there are dependencies in the code already that fail `yarn install` unless running node >= 8.13.0. Then the latest danger-js pushes this to 10.13. I'll try an older version of danger, if possible.

f1337: Here's the PR for the version bump, including the debugging steps performed: https://github.com/department-of-veterans-affairs/vets-json-schema/pull/481

PR #485, opened 2020-08-20
Covid reseach height update

stephenBarrs:
# New schema
Add height as a required field

Coming Soon

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "6.7.2",
+  "version": "7.0.1",
   "repository": {
     "type": "git",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "6.7.2",
+  "version": "7.0.1",
```
mchelen-gov, line 3: does this PR break backwards compatibility? if it's only adding features then a minor version bump should be ok

stephenBarrs: Version is based on the PRs that are in the queue prior to this one @mchelen-gov if this PR is merging prior to those the Version would need to be update as this is a minor or even a patch increment.

PR #487, opened 2020-08-24
Run yarn && yarn build

LinnJS:
# New schema
When opening https://github.com/department-of-veterans-affairs/vets-json-schema/pull/486 @kevinmirc noticed that there were other form changes in my build. So I am opening this PR and just running `yarn && yarn build` so the changes are isolated in another PR. 

### This touches the follow forms:
dist/22-0994-schema.json
dist/21-526EZ-ALLCLAIMS-schema.json

PR #493, opened 2020-08-31
12424 schema for ch36

cohnjesse:
# New schema
[Original Ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/12424)

This PR is to add the schema for the Chapter 36 form. Unit tests for this schema will be included in a separate PR.


src/schemas/28-8832/schema.js
```diff
@@ -1,84 +1,105 @@
...
-      type: 'string',
-    },
-    divorceOrAnnulmentPending: {
-      type: 'boolean',
+    claimantInformation: {
+      type: 'object',
+      properties: {
+        fullName: definitions.fullName,
+        ssn: definitions.ssn,
+        vaFileNumber: definitions.vaFileNumber,
```
micahchiang, line 24: Just to be on the safe side, given all of the recent changes with inflections, it might be worthwhile to change this to `VAFileNumber`.

src/schemas/28-8832/schema.js
```diff
@@ -1,84 +1,105 @@
...
+      properties: {},
     },
-    previousBenefitApplications: {
+    claimantAddress: {
       type: 'object',
       properties: {
-        chapter31: {
-          type: 'boolean',
+        'view:livesOnMilitaryBase': {
+          $ref: '#/definitions/genericTrueFalse',
```
micahchiang, line 36: This needs to be updated since we don't explicitly define any definitions.

src/schemas/28-8832/schema.js
cohnjesse: good idea, updated

src/schemas/28-8832/schema.js
cohnjesse: Again, moving this into a definition is refactoring for refactoring's sake as we are only using this in one place in the schema, however, to keep with tradition I will move it :-)

annaswims:
@cohnjesse Why would you commit changes to dist/21-526EZ-ALLCLAIMS-schema.json in this PR?  they just re-order in a way that isn't helpful.

micahchiang:
@annaswims - looks like those changes may have been part of this prior PR - https://github.com/department-of-veterans-affairs/vets-json-schema/pull/487/files? Not sure why but maybe the merge from master into this branch had some issues 🤷

PR #494, opened 2020-08-31
Fix email minLength

LinnJS:
# New schema
- Remove minLength on email, and update maxLength

## Description

Made this PR due to this [ticket](https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/12585) seemed to be an inconsistency so am fixing in this PR

Updated maxLength based on this [article](https://stackoverflow.com/questions/386294/what-is-the-maximum-length-of-a-valid-email-address#574698)

Frontend:
https://github.com/department-of-veterans-affairs/vets-website/pull/14058

Backend:


LinnJS:
@thilton-oddball the reason I opened this PR in the first place is due to this [ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/12585) I dug deeper and then found that this is inconsistent across VA.gov. The reason I am opening this PR in the first place was because the Frontend validation was inconsistent with other forms.

When @kevinmirc and I pop-wowed we realized there was no real reason for us adding the minLength in the first place and just creates error bloat on the frontend that to your point does not really matter.

Also if it were impossible to have an email this short, which it does seem, why would something be expecting this? Also some more background now that I am thinking about it we added this constant just short while back and not a whole lot of forms are using it other than a handful so it feels that the risk wouldn't really be that high.

I am personally just not convinced "just leaving it the way it is" is necessarily the best way to go about it unless there is a reason it should be 6 chars minLength.

LinnJS:
@kevinmirc and I talked about it and extending the maxLength does make this a `breaking` change and should increment a major version.

PR #495, opened 2020-09-01
Add husky comit hook

LinnJS:
## Description

When opening PR https://github.com/department-of-veterans-affairs/vets-json-schema/pull/494 I found there was unintended changes that had not reached the dist possibly from previous commits. So I opened a PR with just `yarn build` again but added `yarn lint:fix && yarn build` in a husky pre-commit hook to hopefully prevent this kind of thing.

Add husky with pre-commit hooks for build and lint

image


rianfowler:
Can we put the husky changes in a separate PR. I like the idea of adding some linting to this repo but I think we should consider that separately from the schema changes.

rianfowler:
What do we want to do with this PR?

ericboehs:
@LinnJS Any update here? Can we close/move back to draft?

LinnJS:
If we can get some support for this update, I will go through and fix all of the errors currently caught by our rules.

I don't have any idea why people would reject this update. I think its important to enforce rules uniformly for code cleanliness like we do on all our other repos.

image

kelsonic (approved):
Standardization ftw 🚀 We should also add the linter as a check in CI

package.json
```diff
@@ -13,7 +13,16 @@
...
     "lint:fix": "eslint src -c .eslintrc --fix --ext js",
-    "lint:changed": "LIST=`git diff-index --name-only HEAD | grep \"\\.js[x]\\{0,1\\}$\"`; if [ \"$LIST\" ]; then eslint --fix --quiet $LIST; fi"
+    "lint:changed": "LIST=`git diff-index --name-only HEAD | grep \"\\.js[x]\\{0,1\\}$\"`; if [ \"$LIST\" ]; then eslint --fix --quiet $LIST; fi",
+    "prepare": "husky install"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "yarn lint:fix && yarn build"
+    }
+  },
```
theodur, line 23: Do we need this since we already have the `pre-commit` script in the husky directory?

package.json
```diff
@@ -51,5 +62,6 @@
     "**/diff": "^3.5.0",
     "**/growl": "1.10.2",
     "**/minimist": "^1.2.3"
-  }
+  },
+  "dependencies": {}
```
theodur, line 66: Was this added for a specific reason?

package.json
LinnJS: You are right, this is from the old version of Husky

package.json
LinnJS: No idea how this got there will revert, good catches 🙂

LinnJS: Oh looks like this got there from my accidentally adding husky as a `dependency` and not a `devDependency`

LinnJS:
I fixed all ESLint errors other than the ones below since they directly effected the schema `dist` and property names. I feel that these should be fixed by those application owners or each should be done in separate PRs with property references checked on FE and BE on every PR.

image

craigwheeler (approved):

theodur (approved):
This looks good to me! Thanks for making these changes 🚀

LinnJS:
@theodur no probelm just want to make all our lives easier, thanks for being open to the idea 🥇

PR #499, opened 2020-09-08
Update coronavirus-research schema

stephenBarrs:
## Updated COVID-VACCINE-TRIAL-schema

Added namespacing to Select boolean options to distinguish NONE_OF_ABOVE answers
Version updated to `8.2.0`

department-of-veterans-affairs/vets-website#14142


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "8.1.0",
+  "version": "8.2.0",
```
erluti, line 3: would this qualify as a major release (going to `9.0.0`)?

erluti (approved):
LGTM other than a quick question about version

package.json
stephenBarrs: @erluti yes this is a breaking change and major version number should increase.  I will make it so thanks!

PR #505, opened 2020-09-12
Update 10-10CG schema with the form's 10/01/2020 changes

kevinmirc:
**Description**
Update 10-10CG schema with the latest (10/01/2020) data changes.

**Summary**
- [x] Remove insurance questions (medicaidEnrolled, medicareEnrolled, tricareEnrolled, champvaEnrolled, otherHealthInsuranceName).
- [x] Add `primaryCaregiver.hasHealthInsurance` are `required` `boolean`.
- [x] Remove the requirement for `ssnOrTin`s on caregivers


LinnJS (approved):
Looks good all task described in PR are reflected in code change just need to bump the version.

craigwheeler (approved):

bradl3yC (approved):

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "11.0.1",
+  "version": "14.0.0",
```
bkjohnson, line 3: Why is this going from 11 straight to 14?

bkjohnson (commented):
Can you revert the changes to `package-lock.json`? Having that and a yarn lockfile isn't great, so once this gets merged I'll remove `package-lock.json` and have git ignore it in the future.

package-lock.json
kevinmirc: The master branch is on `13.x.x` right now. No one has updated the `package-lock.json` since `11.0.1`, so that's why it's going from 11 to 13.

kevinmirc: @f1337 is there a way to update the GH bot to raise an error when `package.json` version is updated but `package-lock.json` is not? That may help keep them in sync.

bkjohnson: For now we will just remove `package-lock.json` and ignore it from the repo in order to make it clear that the yarn lockfile is the one that gets used.

LinnJS:
@bkjohnson Why would we not want to track our .lock file in git?

package-lock.json
kevinmirc: I've removed the package-lock.json version change.

f1337:
> @bkjohnson Why would we not want to track our `package-lock.json` file in git?

I believe we use yarn.lock for deployments. package-lock.json is ignored outside local environments.

> I am not 100% sure why this is an issue we just merged master and installed...

I think it's more that the update to package-lock.json was unexpected. Not your fault, more an oversight on our part re: we need to add package-lock.json to our .gitignore. IOW, y'all ran across some tech debt.

LinnJS:
Ohhh I get it now, roger that, thanks for the explanation

bkjohnson (approved):

kevinmirc:
Thank you @bkjohnson & @f1337!

PR #522, opened 2020-10-01
20-0996: add `nonPrefill`

edmangimelli:
Additive change to schema:

Adding "nonPrefill" top-level property to `20-0996` schema to allow room for sending extra data to the FE that won't necessarily be included in submission.

wip

### DIFF
--- a/src/schemas/HLR-show-response-200/schema.js
+++ b/src/schemas/HLR-show-response-200/schema.js
@@ -1,4 +1,6 @@
-import schema from '../HLR-create-request-body/schema.js';
+import { cloneDeep } from 'lodash';
+
+const schema = cloneDeep(require('../HLR-create-request-body/schema.js').default);
 
 const definitions = {
   root: {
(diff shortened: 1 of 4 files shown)


src/schemas/HLR-show-response-200/schema.js
```diff
@@ -1,4 +1,6 @@
-import schema from '../HLR-create-request-body/schema.js';
+import { cloneDeep } from 'lodash';
+
+const schema = cloneDeep(require('../HLR-create-request-body/schema.js').default);
```
edmangimelli, line 3: this schema also imported from another schema. cloning as well, out of caution

rianfowler (approved):
Recommend getting peer review on this i.e. someone who understands the context of the change.

edmangimelli:
> Recommend getting peer review on this i.e. someone who understands the context of the change.

@rianfowler thanks! will do

Mottie (approved):

PR #534, opened 2020-10-21
Added income questions

cohnjesse:
# Schema Update
[Original Ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/15141)

This PR is to update the vets-json-schema with new pension questions for the form 686c. These questions are optional and should not be treated as required.


thilton-oddball (approved):
Looks good

PR #540, opened 2020-10-29
[15526] 5655 updates

camerontesterman:
- Adds SSN to schema for veteran identity validation
- Modifies some field names and types to conform to what the DMC is expecting

## Associated PR
https://github.com/department-of-veterans-affairs/vets-api/pull/5187

### DIFF
--- a/dist/5655-schema.json
+++ b/dist/5655-schema.json
@@ -344,128 +344,64 @@
       }
     },
     "income": {
-      "type": "object",
-      "properties": {
-        "veteran": {
-          "type": "object",
-          "properties": {
-            "veteranORSpouse": {
-              "type": "string"
-            },
-            "monthlyGrossSalary": {
-              "type": "string"
-            },
-            "deductions": {
-              "type": "object",
-              "properties": {
-                "taxes": {
-                  "type": "string"
-                },
-                "retirement": {
-                  "type": "string"
-                },
-                "socialSecurity": {
-                  "type": "string"
-                },
-                "otherDeductions": {
-                  "type": "array",
-                  "items": {
-                    "type": "object",
-                    "properties": {
-                      "name": {
-                        "type": "string"
-                      },
-                      "amount": {
-                        "type": "string"
-                      }
+      "type": "array",
+      "items": {
+        "type": "object",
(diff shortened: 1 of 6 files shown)


dist/5655-schema.json
```diff
@@ -261,24 +261,68 @@
...
+    "ssn": {
+      "type": "string",
+      "pattern": "^[0-9]{9}$"
+    },
+    "employmentHistory": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "veteranORSpouse": {
```
ericboehs, line 274:
```suggestion
          "veteranOrSpouse": {
```
Should this be camelCased?

dist/5655-schema.json
```diff
@@ -261,24 +261,68 @@
...
+            "type": "string"
+          },
+          "occupationName": {
+            "type": "string"
+          },
+          "from": {
+            "type": "string"
```
ericboehs, line 281: Is `date` an option here?

dist/5655-schema.json
```diff
@@ -353,40 +349,43 @@
         "veteran": {
           "type": "object",
           "properties": {
+            "veteranORSpouse": {
```
ericboehs, line 352: 🤔

ericboehs (approved):
A few small suggestions but nothing block. LGTM.

dist/5655-schema.json
camerontesterman: nice catch - fixed

dist/5655-schema.json
camerontesterman: Not at the moment as these dates will consist only of month and year and there is still some ongoing discussion about how those will be represented. I will write a regex to capture the correct format once it is decided.

dist/5655-schema.json
camerontesterman: It should be, but our service partners at the DMC have opted to represent it using this format, I am trying to stay consistent with the payload they are ultimately expecting during the submission of the form. I will talk to them about changing it to the format you are suggesting, as it is what I prefer as well, but for now it will have to be like it is.

PR #549, opened 2020-11-19
Unnest medical facility type and name

LinnJS:
# New schema

As of now `lastTreatmentFacility` is an object with a name and type inside like this `lastTreatmentFacility: { name: 'facility name' type: 'clinic' }`. Due to this structure if one is available the other must be present or required. To avoid these two points of data being binded we need to unnest them.

FE Ticket: department-of-veterans-affairs/va.gov-team#15160
BE Ticket: department-of-veterans-affairs/va.gov-team#16477

## Expected Behavior
Unnest the two properties so they are no longer dependent on each other. Currently due to them being in the same object they are both optionally required depending if the other value has data. Make it so they are indigent of each other.

department-of-veterans-affairs/vets-website#15030


LinnJS:
@kevinmirc made this ticket to double check BE department-of-veterans-affairs/va.gov-team#16477

dist/10-10CG-schema.json
```diff
@@ -365,27 +365,17 @@
...
-            },
-            "type": {
-              "type": "string",
-              "enum": [
-                "hospital",
-                "clinic"
-              ]
-            }
-          }
+        "lastTreatmentFacilityName": {
```
LinnJS, line 368: @kevinmirc what do you think about these names?

kevinmirc:
The names look good to me.

dist/10-10CG-schema.json
kevinmirc: The names look good to me.

kevinmirc:
@LinnJS what was the original reason for needing to flatten these properties? I don't see a reference as to why this is needed in any tickets, and I don't recall the reason.

Could we instead just remove the **required** key on the **lastTreatmentFacility** property? That way, you can provide name and type independently, without needing both to be present.

LinnJS:
@kevinmirc I updated the ticket to be more detailed and added the `why` statement in there for you.

kevinmirc:
Just summarizing our convo today about the **required** attribute in the **lastTreatmentFacility** field:

> Since these questions are independent of each other, it is more semantically correct to flatten them rather than keep them nested and optional. We can merge these changes with other major data contract changes since this will require CARMA updates.

kevinmirc (approved):

thilton-oddball (commented):
Will changes also need to be made in vets-api/lib/pdf_fill/forms/va1010cg.rb (https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/pdf_fill/forms/va1010cg.rb) since it references lastTreatmentFacility?

LinnJS:
Created [this ticket](https://app.zenhub.com/workspaces/vsa---caregiver-5fff0cfd1462b6000e320fc7/issues/department-of-veterans-affairs/va.gov-team/21830) to update backed references and assigned to @kevinmirc

LinnJS:
Closing this since I am no longer on this team and its a breaking change and has been open forever.

PR #583, opened 2021-04-08
22853 update parent guardian

cohnjesse:
[Original Ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/22853)

Last sprint we added a question to the 686c asking if the user was the Parent or Guardian of a child they were adding. After stakeholders had a chance to review this it turned out that we actually only care if the user is the Parent, not the guardian, and we should change the question to make that clear. To keep the `formData` consistent with the intent of the form this PR changes the name of the data from `isParentOrGuardian` to `isParent`


micahchiang (approved):
The version number in package.json should be bumped to 20.1.1 I think.

jhonnyoddball (approved):

PR #593, opened 2021-05-24
Added basic schema

cohnjesse:
While developing the dependency verification app it was discovered that we need a schema in vets-json-schema for the form that will be submitted, the 21-0538. This PR adds that schema.


src/schemas/21-0538/schema.js
```diff
@@ -0,0 +1,34 @@
+import cloneDeep from 'lodash/cloneDeep';
+import pick from 'lodash/pick';
+import commonDefinitions from '../../common/definitions';
+
+const definitions = cloneDeep(commonDefinitions);
+
+const schema = {
+  $schema: 'http://json-schema.org/draft-04/schema#',
+  title: 'DEPENDENCY VERIFICAION (0538)',
```
micahchiang, line 9: Verificaion should be Verification

micahchiang (commented):
Not sure how it happened but this PR includes two new schemas in `/dist` instead of just one.

src/schemas/21-0538/schema.js
cohnjesse: strange, I removed the extra one

cbtkc (approved):
Looks good!

jhonnyoddball (approved):

PR #602, opened 2021-06-04
Changed names of document upload

cohnjesse:
[Original Ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/25791)

During testing of the 686c form it was discovered that if a user went through two separate workflows and uploaded different documents for each workflow they were only allowed to upload one type of document instead of two separate ones. This turned out to be caused by both document uploaders using the same array so this PR separates the arrays that are used for the document uploads on the 686c.

### DIFF
--- a/dist/686C-674-schema.json
+++ b/dist/686C-674-schema.json
@@ -1507,7 +1507,7 @@
                 "Unknown"
               ]
             },
-            "supportingDocuments": {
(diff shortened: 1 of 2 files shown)


timwright12 (approved):

micahchiang (commented):
I think this PR needs a version bump to the vets-json-schema package.json

PR #611, opened 2021-06-25
526: Update treatment date range requirement

Mottie:
## Description

The newest 526 form removes the requirement for a VA treatment start date & does not include a treatment end date. This schema update removes the requirement, but maintains the structure  to maintain backward compatibility.

See https://github.com/department-of-veterans-affairs/va.gov-team/issues/26265

## Pull Requests to update the schema in related repositories

https://github.com/department-of-veterans-affairs/vets-website/pull/17694

PR #612, opened 2021-06-30
update HLR schemas from Lighthouse

annaswims:
# New schema
No changes that affect production code, but new comments from Lighthouse's schema updates.

## Pull Requests to update the schema in related repositories
The changes are mostly additional comments so I'm not going to bump the versions of vets-api vets-website It can just be included in the next update

PR #613, opened 2021-06-30
add file that was missed as part of #611

annaswims:
this was missing from #611

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "20.6.2",
+  "version": "20.6.4",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/department-of-veterans-affairs/vets-json-schema.git"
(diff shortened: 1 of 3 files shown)


timwright12:
@annaswims Looks like there are some failing tests related to this update

annaswims:
thanks, @timwright12! I deleted the test that was no longer necessary as well

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "20.6.2",
+  "version": "20.6.4",
```
Mottie, line 3: version bumped twice 🤷🏽‍♂️

Mottie (approved):
ghe-emoji

package.json
annaswims: I also bumped the version in https://github.com/department-of-veterans-affairs/vets-json-schema/pull/612

ericboehs (approved):

timwright12:
Going to defer the review to @ericboehs as I'm not familiar with the code and it looks like you can merge without me

PR #620, opened 2021-07-19
26245 remove enum constraint for 1010CG facility field

dnalbach:
# New schema

department-of-veterans-affairs/va.gov-team#26245

None required as this only removes a constraint rather than adding one. This is in preparation for pulling a live API response of the Caregiver facility list instead of hardcoding it in the schema. Removing the enum constraint will allow both the existing code flow and the future one (being developed under feature toggle) to each pass.


jhonnyoddball (approved):
LGTM

rileyanderson (approved):

PR #622, opened 2021-07-20
Added schema for COE

cohnjesse:
# Background
Now that the form for COE is relatively stable from a design standpoint, we should move the temporary schema we have been using in vets-website to a permanent schema in vets-json-schema. This PR does just that.


jhonnyoddball (approved):
LGTM

k80bowman (approved):

PR #661, opened 2021-11-10
Add SIGI Self Identifying Genders to 10-10ez schema

LinnJS:
## Updates
- [x] Add sigiGenders to constants
- [x] Add sigiGender to 10-10ez schema

department-of-veterans-affairs/va.gov-team#25411

department-of-veterans-affairs/vets-website#19337


package.json
```diff
@@ -52,4 +52,4 @@
     "**/growl": "1.10.2",
     "**/minimist": "^1.2.3"
   }
-}
+}
```
theodur, line 55: It looks like the newline was removed in this file. Can we add that back in here?

LinnJS: @theodur if this is a rule we are going enforce should really be a ESLint rule, this is auto-formatting at work.

LinnJS: Does not look like we enforce ESLint rules at all...

image

LinnJS: Here is a [throwback PR](https://github.com/department-of-veterans-affairs/vets-json-schema/pull/495), we should really enforce our own rules, otherwise what is the point of having them?

theodur: Didn't realize the rules weren't being enforced in the repo. If we can get that PR working, that'll be great to merge in. I'll take a look it.

theodur (approved):

craigwheeler (approved):

PR #685, opened 2022-03-11
28-1900 schema: update appointmentTimePreferences from object to array

RudyOnRails:
# New schema

This PR is to modify the schema for the Veteran Readiness and Employment VA Form 28-1900 (Chapter 31). In production we are currently running into database update errors because we are passing in an array, and the schema expects an object.

`app/models/saved_claim/veteran_readiness_employment_claim.rb`

```
(byebug) update!(form: form_copy.to_json)
*** ActiveRecord::RecordInvalid Exception: Validation failed: Form The property '#/appointmentTimePreferences' of type array did not match the following type: object in schema 725bb504-ad14-510a-8dd6-6782b26e7cf5
```

## Related tickets

- http://sentry.vfs.va.gov/organizations/vsp/issues/49373/?environment=production&project=3&statsPeriod=90d#exception
- https://github.com/department-of-veterans-affairs/va.gov-team/issues/36744

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "20.19.4",
+  "version": "21.0.0",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/department-of-veterans-affairs/vets-json-schema.git"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "20.19.3",
+  "version": "20.19.4",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/department-of-veterans-affairs/vets-json-schema.git"
--- a/dist/28-1900-schema.json
+++ b/dist/28-1900-schema.json
@@ -608,18 +608,21 @@
     },
     "appointmentTimePreferences": {
       "type": "array",
-      "properties": {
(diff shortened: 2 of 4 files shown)


timwright12 (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vets-json-schema",
-  "version": "20.19.3",
+  "version": "21.0.0",
```
jerekshoe, line 3: This should probably be `20.19.4` I think.

RudyOnRails: Ok, I was going with what I read here...

https://github.com/department-of-veterans-affairs/vets-json-schema#:~:text=Changing%20the%20type%20on%20a%20schema%27s%20property

jerekshoe: @timwright12 can you shed some light on this? I just want to check for my own future reference

timwright12: I don't know enough about this data to really say whether it's a breaking change or not. @RudyOnRails would you be able to look at where it's used/implemented to confirm a full version bump?

RudyOnRails: @timwright12 Regardless of the data, I was strictly following the readme guidelines of what was considered a major bump:
> Changing the type on a schema's property

RudyOnRails:
Found out the problem laid on the backend...

PR #697, opened 2022-04-19
add pst-forms-library to codeowners

micahchiang:
## Description
Since the Forms Library Platform Spike Team has taken over support for the forms library from the vsp design system team, it needs to be added as a codeowner for things related to the forms library. [This ticket includes an ongoing conversation for historical context](https://github.com/department-of-veterans-affairs/va.gov-team/issues/33037).

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,4 @@
 # These owners will be the default owners for everything in
 # the repo. Unless a later match takes precedence,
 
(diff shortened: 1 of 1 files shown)


ericboehs:
Looks like invalid syntax (missing team?). Moving to draft.

micahchiang:
@ericboehs - yeah sorry about that. I guess the frontend-review-group team was removed from the org? I've updated the PR and the codeowners file should be valid now 👍

ericboehs (approved):

PR #711, opened 2022-07-29
1010ez state fix

lihanli:
# New schema

update 1010ez schema to only allow 50 US states + DC + "Other"
https://github.com/department-of-veterans-affairs/va.gov-team/issues/45079


lihanli:
@longmd not sure

longmd:
@lihanli , I suppose I could have done some reading before asking... 

`Breaking changes include: Removing a property from a schema, Adding a property to a schema that has additionalProperties set to false, Making a property required that was not previously required on that schema, Changing the type on a schema's property, Removing values in the enum key of a schema's property, ect.`

Looks like by removing those values, that is a breaking change.

jeremyhwu (approved):

PR #726, opened 2022-10-21
Removing service branches enum from 526 schema

data-doge:
(PR #725 was reverted since more frontend (FE) work was needed. Having this change in place would likely break new 526 submissions without the FE work completed. Synchronizing updates of this change with the FE & backend are essential.)

In https://github.com/department-of-veterans-affairs/va.gov-team/issues/41046 we are beginning to consume lighthouse's /service-branches endpoint (See https://developer.va.gov/explore/benefits/docs/benefits_reference_data?version=current) to populate our dropdowns for military service branches in the 526ez form. Since we now expect that branches may be added or removed in the future, we can no longer enforce an enum.

# New schema

department-of-veterans-affairs/va.gov-team#41046

https://github.com/department-of-veterans-affairs/vets-website/pull/22476
department-of-veterans-affairs/vets-api#0000 (N/A, no backend changes)

### DIFF
--- a/dist/21-526EZ-ALLCLAIMS-schema.json
+++ b/dist/21-526EZ-ALLCLAIMS-schema.json
@@ -847,16 +847,7 @@
       }
     },
     "militaryRetiredPayBranch": {
-      "type": "string",
-      "enum": [
-        "Air Force",
-        "Army",
-        "Coast Guard",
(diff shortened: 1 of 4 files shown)


Mottie:
Holding off on merging until vets-website code is ready for this change

PR #753, opened 2023-03-10
Add json schema + tests for 21-4142

rhasselle-oddball:
# New schema
For form 21-4142 for release and disclosure of medical records to the VA
- department-of-veterans-affairs/va.gov-team-forms#89

Notes on old PDF vs new PDF for `form4142`
The `providerFacility` property is similar to `#/definitions/form4142` (2018 version) with some slight differences to match the 2021 PDF:
- added `conditionsTreated` property 
- `treatmentDateRange` is no longer an array, but an object, since there is only one set of value in the new PDF
- `privacyAgreementAccepted` should be for the entire form 4142/4142a rather than just the definition, so pulled that out to the correct scope

For all these reasons, I didn't reuse the "common" definition `form4142`.

After you've merged your changes to vets-json-schema you'll need to make PR's to vets-website and vets-api. Please link them here.
- [ ] department-of-veterans-affairs/vets-website#0000
- [ ] department-of-veterans-affairs/vets-api#0000

### DIFF
--- a/dist/21-4142-schema.json
+++ b/dist/21-4142-schema.json
@@ -1,0 +1,909 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Authorize the release of medical information to the VA",
+  "type": "object",
+  "definitions": {
+    "address": {
+      "type": "object",
+      "oneOf": [
+        {
+          "properties": {
+            "country": {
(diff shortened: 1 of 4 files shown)


considerable (approved):

rmessina1010 (approved):
LGTM

PR #861, opened 2024-03-14
Revert "Rename form 28-8832 to 27-8832"

zack:
We went about this all the wrong way.

This reverts commits:
- 8fe4f0622ac13ab5105bec76f050779b315363b1
- 81fe1863f7280db3640138888900d1d59cfa4ecc


dist/version.json
```diff
@@ -1,3 +1,3 @@
 {
-  "version": "21.2.0"
+  "version": "21.1.4"
```
zack, line 2: Not sure if we should revert this or bump it again?

LindseySaari: I think we would have to "fail foward"

zack: 👍

zack:
Thanks @LindseySaari. Sorry for all the noise today and yesterday. :(

dist/version.json
```diff
@@ -1,3 +1,3 @@
 {
-  "version": "21.2.0"
+  "version": "21.3.0"
```
LindseySaari, line 2: Does the version and package.json need to be the same?

tblackwe (approved):

zack:
Hi @LindseySaari, is everything in order? We'd like to merge this to unblock another team. Thanks!

rmessina1010 (approved):
LGTM

PR #871, opened 2024-03-28
Support local bundle gemspec

scottsdevelopment:
# Support local bundle gemspec

Add absolute pathing to the gemspec file, this helps with a bug in local bundle configuration about not finding a file.


mjknight50 (approved):
LGTM

stevenjcumming (approved):

PR #884, opened 2024-04-16
Local development linking

scottsdevelopment:
- Add commands for local development linking
- Update documentation on commands
- Remove version.json as it won't be used
- Add license to package json (inferred by vets-api and vets-website) to prevent warning spam

- department-of-veterans-affairs/va.gov-team#79995

### DIFF
--- a/script/local-link.sh
+++ b/script/local-link.sh
@@ -56,7 +56,7 @@ unlink_vets_website() {
     echo "Unlinking vets-website..."
     cd ../vets-website
     yarn unlink vets-json-schema
-    # yarn -W add vets-json-schema
+    yarn install
     cd ../vets-json-schema
     yarn unlink
     echo "vets-website unlinked."


package.json
```diff
@@ -11,9 +12,15 @@
     "test": "mocha 'test/**/*.spec.js'",
     "test:all-schemas": "mocha 'test/common/all-schemas.spec.js'",
     "watch": "node src/start watch",
+    "link": "yarn link:vets-api; yarn link:vets-website",
```
wayne-weibel, line 15: use a different name so there is no confusion with the builtin yarn command

package.json
```diff
@@ -11,9 +12,15 @@
...
+    "link:vets-api": "cd ../vets-api; bundle config set local.vets_json_schema ../vets-json-schema; bundle config disable_local_branch_check true",
+    "link:vets-website": "yarn link; cd ../vets-website; yarn link vets-json-schema",
```
wayne-weibel, line 17: should move these to a shell script, to keep the package.json cleaner and can annotate the commands

wayne-weibel (changes requested):
Please add a way to test that the local json-schema is being referenced ...

I am seeing an ‘error’ doing unlink and the script is not using the command noted in the output to reinstall the package; would also like this to be in a shell script so it it easier to read and can be annotated

image

wayne-weibel (approved):

RachalCassity (approved):

pmclaren19:
@RachalCassity @scottsdevelopment @wayne-weibel trying to follow these instructions here and they are not working. Anyone free to pair?

PR #885, opened 2024-04-26
Fix gemspec

nihil2501:
I wasn't able to `bundle install` `vets-api` after a `bundle pristine`:
```
> bundle install
Fetching gem metadata from https://rubygems.org/.......
Fetching gem metadata from https://enterprise.contribsys.com/..
Your bundle is locked to vets_json_schema (22.1.1) from https://github.com/department-of-veterans-affairs/vets-json-schema (at master@7176cb0), but that version can no longer be found in that source. That means the author of
vets_json_schema (22.1.1) has removed it. You'll need to update your bundle to a version other than vets_json_schema (22.1.1) that hasn't been removed in order to install.
```
I saw [this PR](https://github.com/department-of-veterans-affairs/vets-json-schema/pull/871) which changed `vets_json_schema.gemspec`. Making this branch that reverts that part of the change and updating my `vets-api` `Gemfile` and `Gemfile.lock` to reflect it fixed the issue for me:
```
GIT
  remote: https://github.com/department-of-veterans-affairs/vets-json-schema
  revision: b13126b2840636be516e652a501e0d4870c53a23
  branch: fix-gemspec
  specs:
    vets_json_schema (22.1.1)
      multi_json (~> 1.0)
      script_utils (= 0.0.4)
```
Feel free to take over the work from me though since I don't know what the other issue around local bundling was about!


vets_json_schema.gemspec
```diff
@@ -12,4 +12,4 @@ Gem::Specification.new do |gem|
 
   gem.add_dependency 'multi_json', '~> 1.0'
   gem.add_dependency 'script_utils', '0.0.4'
-end
```
RachalCassity, line 15: @nihil2501 Can you please add the empty line back to this file?

nihil2501: @RachalCassity I undid that edit.

RachalCassity (approved):

nihil2501:
> According to documentation and [code](https://github.com/rubygems/bundler/blob/master/lib/bundler/cli/pristine.rb#L31), `bundle pristine` will remove the packages when the source is remote, and then force a checkout. Since the `Gemfile` defines packages in the form of `gem 'vets_json_schema', git: 'https://github.com/department-of-veterans-affairs/vets-json-schema'`, the latest version of the `master` will be checked out, this file will not contain the version of the package that is being requested in a `Gemfile.lock` and cause this error.
> 
> The solution here is to call `gem update vets_json_schema` or `gem update bgs_ext` (in my case) so that the lock file may be updated first before doing pristine install. Alternatively, we could target specific commit hashes inside of our Gemfile moving forward.

@scottsdevelopment Thanks for outlining, how this works. TIL! Sorry about the human error here!

PR #893, opened 2024-05-03
EZR: Reject whitespace on required input fields

dellerbie:
# New schema
This PR updates the schema for the 10-10EZ & 10-10EZR to reject whitespace on required text fields.

## Related issue(s)
- https://github.com/department-of-veterans-affairs/va.gov-team/issues/78895

_After you've merged your changes to vets-json-schema you'll need to make PR's to vets-website and vets-api. Please link them here._

- https://github.com/department-of-veterans-affairs/vets-api/pull/
- https://github.com/department-of-veterans-affairs/vets-website/pull/

### DIFF
--- a/test/common/definitions.spec.js
+++ b/test/common/definitions.spec.js
@@ -464,6 +464,17 @@ describe('schema definitions', () => {
         country: '     ',
         provinceCode: '     ',
       },
+      // doesn't allow postalCode with only spaces
+      {
+        country: 'USA',
+        state: 'AK',
+        street: stringGenerate(23),
+        street2: stringGenerate(13),
+        street3: stringGenerate(5),
+        city: stringGenerate(27),
+        postalCode: stringGenerate(5),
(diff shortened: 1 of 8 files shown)


mkerns1 (approved):
LGTM! Nice work on this Derrick!

test/common/definitions.spec.js
```diff
@@ -463,6 +463,7 @@ describe('schema definitions', () => {
         city: '    ',
         country: '     ',
         provinceCode: '     ',
+        postalCode: '    ',
```
Mitch-A6, line 466: 🤔 Does adding this to the test change whether it passes or not? Conversely, would this still fail if we removed the code it should be accounting for? I'm not sure how specific we tend to get with these tests, but from a TDD perspective this feels like it wouldn't add to coverage - the invalid schema is still invalid with the newly invalid field.

RachalCassity (approved):

PR #1001, opened 2025-05-08
Adds two additional options for the private medical records dropdown

tyler-spangler6:
# New schema
This adds additional options to the schema for the 21-526EZ-allclaims for the field `privateMedicalRecords`.  The additional options will allow Veterans to have more specific document types when they upload their Private Medical records.  

- https://github.com/department-of-veterans-affairs/abd-vro/issues/4315

After this is merged, will update vets-website and vets-api with the corresponding commit hash to ensure all repos are up to date.


dist/21-526EZ-ALLCLAIMS-schema.json
```diff
@@ -1773,10 +1773,14 @@
           "attachmentId": {
             "type": "string",
             "enum": [
+              "L702",
+              "L042",
```
dfitchett, line 1777: `L042` should be `'L049' => 'Medical Treatment Record - Non-Government Facility'` in the [lighthouse document API](https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/models/lighthouse_document.rb). This needs to be fixed in the schema.js file too!

dfitchett (changes requested):

dist/21-526EZ-ALLCLAIMS-schema.json
tyler-spangler6: Thanks, sorry about that typo! It should be fixed now.

dfitchett (approved):

dfitchett (approved):

rjohnson2011 (approved):

tommasina-va (approved):
LGTM!

etanb:
Everything looks good from an accessibility stand point. Only have one question:

How was ordering determined? Usually the default ordering is "alphabetical", but was there another assumption in place here like sorting by frequency? IE, `Medical Treatment Record - Non-Government Facility` is chosen more frequently than `Disability Benefits Questionnaire (DBQ)`?

tyler-spangler6:
@etanb Thanks for reviewing this Etan. That is correct; it is ordered by frequency of selection.

